PR #5, opened 2016-12-04
Add types to support heterogeneous collections of events.

jimmycuadra:
Some of the Matrix APIs involve JSON arrays containing a heterogeneous collection of events. This doesn't play well with static typing, specifically with the way the events are designed in ruma-events.

For example, the [event context API](https://matrix.org/docs/spec/client_server/r0.2.0.html#get-matrix-client-r0-rooms-roomid-context-eventid) includes lists of room events and state events in its response body. To represent this in Rust, we'd use something like:

``` rust
Vec<Box<ruma_events::RoomEvent>>
```

However, the `RoomEvent` trait (and its counterparts) includes an associated type that declares the type used for the event's `content` field. This means that it needs to be declared like this:

``` rust
Vec<Box<ruma_events::RoomEvent<Content=???>>>
```

Unfortunately, there isn't a way to use a generic to constrain the associated type in a way that wouldn't either force the whole vector to be events of only one concrete type or use a less specific type like `serde_json::Value` as the content type, which would defeat the purpose of having static types for each event type.

This branch adds a few enums which represent different subsets of the various event types to support heterogeneous collections like this. With these new enums, the vector would be represented as:

``` rust
Vec<Box<ruma_events::collections::OnlyRoomEvent>>
```

I'm not thrilled about the naming scheme of these enums, since the above code really represents a list of "room events," not a list of "only room events." But of course, `RoomEvent`, et. al, is already taken by the traits.

Looking for feedback/ideas on this approach, possible alternate approaches, and the names I'm using.

/cc @ruma/contributors


farodin91:
You could extend this with Into<Enum> like https://github.com/ruma/ruma/issues/115. This would make it easier to use.

mujx:
I like the approach with the enums, it seems pretty straightforward. The naming part indeed seems a little off. Instead of trying to come up with the most generic names maybe we could adjust the collections specifically for each endpoint and generalize where it makes sense.

The events of the context api could be named something like `collections::ContextEventsAfter`, `collections::ContextEventsBefore` or `collections::ContextStateEvents`.

jimmycuadra:
That's an interesting idea. I'm betting we'll end up with several types that are functionally equivalent by doing that. It'd also make less sense to have these collection types in ruma-events if they were named for specific APIs, since ruma-events is supposed to be agnostic to how it's used.

Another possibility would be actually calling the collection types `Event`, `RoomEvent`, and `StateEvent`, but distinguishing them based on module. In other words, use a path like `collections::any::RoomEvent` or `collections::only::RoomEvent` at the point of usage.

jimmycuadra:
What do you think of the latest changes?

farodin91:
Changing from into a macro to reduce same code?
```rust
impl From<TypingEvent> for Event {
    fn from(typing_event: TypingEvent) -> Self {
        Event::Typing(typing_event)
    }
}
```
into
```rust
from_impl!(TypingEvent,Event,Typing);
```

jimmycuadra:
Closed by c92ce9eeb1fe53e62840e4b55ba55b0538d771ed.

PR #7, opened 2016-12-17
Add missing stripped StateEvents

mujx:
Fixes deserialization issues from ruma/ruma#127

### DIFF
--- a/src/stripped.rs
+++ b/src/stripped.rs
@@ -1,27 +1,60 @@
 //! Stripped-down versions of certain state events.
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Error as SerdeError, Serialize, Serializer};
+use serde_json::{Value, from_value};
 
 use EventType;
+use room::aliases::AliasesEventContent;
 use room::avatar::AvatarEventContent;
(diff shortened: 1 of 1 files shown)


src/stripped.rs
```diff
@@ -1,27 +1,60 @@
...
 
 use EventType;
+use room::aliases::AliasesEventContent;
 use room::avatar::AvatarEventContent;
 use room::canonical_alias::CanonicalAliasEventContent;
-use room::join_rules::JoinRulesEventContent;
+use room::create::CreateEventContent;
+use room::guest_access::GuestAccessEventContent;
+use room::history_visibility::HistoryVisibilityEventContent;
+use room::join_rules::*;
```
jimmycuadra: Use explicit imports here, please!

jimmycuadra:
Looks good, just one minor change to get rid of the glob import. Thanks!

mujx:
I used the glob to avoid the unused  import warning for `JoinRule` but I fixed it with `super::super`.

jimmycuadra:
I'd just use the regular import path in the tests rather than `super`, for items that aren't actually defined in the parent module. That is, only `StrippedRoomTopic` and `StrippedState` would be imported using `super`.

PR #8, opened 2017-01-05
Add missing trait implementations for PresenceState

mujx:
Came up while reviewing the presence endpoints.


jimmycuadra:
Good catch, thanks!

PR #9, opened 2017-01-09
Update matrix events to their latest version

mujx:
A work in progress so we can integrate the latest changes. 

So far it matches the [HEAD](https://matrix.org/speculator/spec/HEAD/client_server/unstable.html) of the spec with the exception of the [new](https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#m-direct) event `m.direct`. It will be updated (if any change occurs) until `v0.3.0` comes out.


mujx:
Now the PR should be up-to-date with the latest spec version (`v0.3.0`)

mujx:
No worries, I'm glad you're back working on ruma.

PR #12, opened 2017-04-27
Power level defaults

exul:
http://matrix.org/docs/spec/client_server/r0.2.0.html#m-room-power-levels describes three fields that have default values if they are not specified, this PR adds those default values.


farodin91:
`ban`, `invite`, `redact` and `kick` should be also 50.

farodin91:
Thanks

PR #15, opened 2017-08-03
More bytesize fixes

jplatte:
Links to relevant spec parts:

* `m.call.*`: https://matrix.org/docs/spec/client_server/r0.2.0.html#voice-over-ip
* `m.receipt`: https://matrix.org/docs/spec/client_server/r0.2.0.html#m-receipt
* `m.room.avatar`: https://matrix.org/docs/spec/client_server/r0.2.0.html#m-room-avatar

### DIFF
--- a/src/room/mod.rs
+++ b/src/room/mod.rs
@@ -21,11 +21,11 @@ pub mod topic;
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct ImageInfo {
     /// The height of the image in pixels.
-    pub height: u64,
+    pub h: u64,
     /// The MIME type of the image, e.g. "image/png."
     pub mimetype: String,
     /// The file size of the image in bytes.
     pub size: u64,
     /// The width of the image in pixels.
-    pub width: u64,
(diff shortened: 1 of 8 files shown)


src/room/mod.rs
```diff
@@ -21,11 +21,11 @@ pub mod topic;
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct ImageInfo {
     /// The height of the image in pixels.
-    pub height: u64,
+    pub h: u64,
```
jimmycuadra: Let's leave the field as `height` but do a serde rename for serialization.

src/room/mod.rs
```diff
@@ -21,11 +21,11 @@ pub mod topic;
...
     /// The MIME type of the image, e.g. "image/png."
     pub mimetype: String,
     /// The file size of the image in bytes.
     pub size: u64,
     /// The width of the image in pixels.
-    pub width: u64,
+    pub w: u64,
```
jimmycuadra: Same here, leave as `width`.

PR #16, opened 2017-12-06
Fix typing notification content struct

rschulman:
The spec only indicates that this field should be a string, but makes clear in the description and example that the content should be a full user ID, not an event ID. This PR fixes the struct so it can deserialize properly.


mujx:
This is already handled by #9 btw (in case you need an up-to-date version of the structs).

jimmycuadra:
I'm gonna close this since it's included in mujx's PR, but thank you, Ross!

PR #20, opened 2018-12-25
Make the room_id optional in sync-related events

### LINKED ISSUE #19: `room_id` field should not be in room events of /sync response
As hinted at in ruma/ruma-client-api#27 the room events in the response to /sync do not have the usual `roomId` field resulting in ``RumaApi(SerdeJson(Error("missing field `room_id`")))``.

Looking at the unstable spec, this is even explicitly noted now (was just absent before):
>  room_id The ID of the room associated with this event. Will not be present on events that arrive through /sync, despite being required everywhere else.

vberger:
According to the spec, this field can be absent in some contextes,
notably in the responses to the /sync endpoint, where the events
appear in a context where there is no ambuity regarding which room
the event is related to.

Fixes #19 and makes ruma-client now capable of parsing /sync responses
from synapse 0.34.0


vberger:
Fixed.

I also made optional some other fields that can be missing from the summarized events in an initial sync.

vberger:
I reverted the other changes to only have the `room_id` ones.

It appears the issue with the others is that the `invite_state` in a sync response should contain stripped state events, and ruma-client-api does not respect that.

PR #30, opened 2019-09-29
[WIP] EventResult API

jplatte:
Still left to do:

* Implement Serialize for the raw event types that don't implement it yet
* ~~Figure out what to to with the orphan `impl TryFrom` for StrippedEventContent~~
  * ~~One possibility would be to replace usage of `TryFrom` in `EventResultCompatible` with a custom conversion function~~ I have implemented this
* Implement the remaining `unimplemented!()` functions
* Adjust documentation


src/stripped.rs
```diff
@@ -81,80 +80,79 @@ pub struct StrippedStateContent<C> {
...
 
+/*impl<C> EventResultCompatible for StrippedStateContent<C>
+where
+    C: EventResultCompatible,
+{
+    type Raw = StrippedStateContent<C::Raw>;
+}
+
+// Orphan impl :(
+impl<C, E> TryFrom<StrippedStateContent<C::Raw>> for StrippedStateContent<C>
```
jplatte: @jimmycuadra This is the orphan impl I referenced in the PR description

src/lib.rs
```diff
@@ -236,6 +245,105 @@ impl Display for InvalidInput {
...
+}
+
+/// The result of deserializing an event, which may or may not be valid.
+///
+/// When data is successfully deserialized and validated, this structure will contain the
+/// deserialized value `T`. When deserialization succeeds, but the event is invalid for any reason,
+/// this structure will contain an `InvalidEvent`. See the documentation for `InvalidEvent` for
+/// more details.
+#[derive(Clone, Debug)]
+pub enum EventResult<T: EventResultCompatible> {
```
florianjacob: I don't see why the enum is needed, would a simple type alias for Result not be sufficient? Especially as you're able to tranpsarently convert it to a Result-based type, and it's surprising that something called `EventResult` is *not actually* a type alias, as with the usual `IoResult<T>` and similar types which fix the Error type, like the common `std::io::Result as IoResult`.

Something like:
`type EventResult<T> = Result<T, InvalidEvent<T::Raw>;`

ruma-events-macros/tests/ruma_events_macros.rs
```diff
@@ -0,0 +1,373 @@
...
+            }
+        }
+
+        deserializer.deserialize_str(EventTypeVisitor)
+    }
+}
+
+/// The result of deserializing an event, which may or may not be valid.
+#[derive(Debug)]
+pub enum EventResult<T: EventResultCompatible> {
```
florianjacob: There are two identical definitions of EventResult except the derives, one in ruma-events-macros, one in ruma-events. Does this make sense?

ruma-events-macros/tests/ruma_events_macros.rs
```diff
@@ -0,0 +1,373 @@
...
+        }
+    }
+}
+
+pub trait EventResultCompatible {
+    /// The raw form of this event that deserialization falls back to if deserializing `Self` fails.
+    type Raw;
+    type Err: Into<String>;
+
+    fn try_from_raw(_: Self::Raw) -> Result<Self, (Self::Err, Self::Raw)>;
```
florianjacob: It feels weird to call something `EventResultCompatible` which has a function which does not return an `EventResult` but an actualy `std::Result`.

ruma-events-macros/tests/ruma_events_macros.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Convert `EventResult<T>` into the equivalent `std::result::Result<T, InvalidEvent<T::Raw>>`.
+    pub fn into_result(self) -> Result<T, InvalidEvent<T::Raw>> {
+        match self {
+            EventResult::Ok(t) => Ok(t),
+            EventResult::Err(invalid_event) => Err(invalid_event),
+        }
+    }
+}
+
+pub trait EventResultCompatible {
```
florianjacob: From the name alone, I would think this is a compatibility adaptor, e.g. between old and new API versions.
Wouldn't it be possible to use the std's `TryFrom`, e.g. with a type alias? Especially with the `Void` type, this looks to me like a reimplementation of `TryFrom`/`Into` and `infallible`.

If that is not possible for some reason I don't see yet, that reason should be documented (future rust might be able to do it), and I would still consider to go with the usual Rust convention and name it something like `FromRaw` or `FromRawEvent` I'd guess.

src/lib.rs
```diff
@@ -776,6 +881,35 @@ where
...
+            .collect::<HashMap<_, _>>()
+            .serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
+    where
+        D: Deserializer<'de>,
+        T: Deserialize<'de> + Hash + Eq,
+    {
+        HashMap::<T, Empty>::deserialize(deserializer)
```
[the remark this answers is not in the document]
jplatte: It could fail deserialization for sure, but why do you think it could panic? Either way, this is not something this PR changes, I just changed implementation details. But feel free to open an issue about `Empty` deserialization :)

src/lib.rs
jplatte: It's required because we need to implement `Deserialize` for it, which we wouldn't be allowed to otherwise (and in fact serde already contains a`Deserialize` implementation for `std::result::Result`).

jplatte:
I'll answer you in more detail tomorrow, but let me just say thanks for taking the time to review this! 👍

All of the API-related things you commented on are intentional, but naming and docs certainly need to be improved before releasing a new version.

src/lib.rs
```diff
@@ -776,6 +881,35 @@ where
     }
 }
 
+/// Serde serialization and deserialization functions that map a `Vec` to a
```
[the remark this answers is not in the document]
jplatte: :+1:

ruma-events-macros/tests/ruma_events_macros.rs
jplatte: If you have ideas for better names for `EventResult` & `EventResultCompatible`, I'm all ears! Those are basically just the first things that came to mind when this API was designed.

ruma-events-macros/tests/ruma_events_macros.rs
jplatte: >  Wouldn't it be possible to use the std's `TryFrom`

An earlier revision of this did use `TryFrom` (or actually `TryInto` I think). However, that didn't work out due to coherence rules (see https://github.com/ruma/ruma-events/pull/30/commits/c20d79db7d44f4422dacc0962ddcfb6d5e65d550#diff-dac0877fbdc2ec814721cb8870d4f8ccL138).

> Especially with the `Void` type, this looks to me like a reimplementation of `TryFrom`/`Into` and `infallible`.

It is. It's not really meant to be public though, I'll have to look into making all of this private implementation details. I explained above why we can't use `TryFrom`. I actually thought of using `Infallible` too, but that doesn't work because there is no `impl From<Infallible> for String`. This will be solved once `!` is stable, because there is an `impl<T> From<!> for T`. Then we'll change `Void` to `!`. (I've recently added a comment to the `Void` type about that).

> I would still consider to go with the usual Rust convention and name it something like `FromRaw` or `FromRawEvent` I'd guess.

Thanks for the inspiration! `TryFromRaw` could be a good new name for `EventResultCompatible`. What do you think?

src/lib.rs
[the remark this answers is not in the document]
florianjacob: You're right, wrongly linked deserialization failure to panicking (too much unwrapping in my client code), and it's indeed off-topic here. :) When the ruma-events dust has settled, I'll go and add a unit test for what I would think is the correct behaviour and see what happens.

ruma-events-macros/tests/ruma_events_macros.rs
florianjacob: Yeah, if it's just plain reorganization of ruma-events-macros into ruma-events then let it be for now, PR is large enough as it is.

src/lib.rs
florianjacob: I missed the Deserialize impl, probably looked at the ruma-events-macros definition instead.
The usage of `EventResult` in [ruma-client-api](https://github.com/ruma/ruma-client-api/commit/98e0bad1b1f53f6376c3bc3a8bc752f1b54b35bb) completes the picture why `Deserialize` is needed here and how `EventResult` is used. I now don't see any other option than reimplementing Result as well, and would go with that option for now. When this is finished, it might make sense to go and ask the serde people whether there's some clever solution to this, [like it has happened before](https://github.com/ruma/ruma-events/pull/25#discussion_r255252248).

Would it be possible to implement `From/Into<std::result::Result<T, InvalidEvent<T::Raw>>>` for `EventResult`? Could make the API nicer and the conversion more transparent in some places, or am I missing something again why that also can't be implemented? xD

ruma-events-macros/tests/ruma_events_macros.rs
florianjacob: Unfortunate, but if we can hide it away from the public interface it's a fine workaround.

On Naming: Now that I have the full picture, I would go with `EventResult` and try to make the difference to `std::Result` as transparent as possible, and `TryFromRaw`, which maches the convention to name the trait like the main / single method.

src/lib.rs
jplatte: I'm not entirely sure. I think we'll be able to implement it, will try later.

jplatte:
@ruma/client I need one approving review to merge this.

jplatte:
Merged (manually)! :)

PR #36, opened 2019-11-19
Replace FromRaw with Deserialize and remove corresponding "raw" modules

lu-fennell:
Hi. I was very inspired by your live-coding in August on the events, and I played around with the code a bit. I think I found a slight simplification to the FromRaw approach:

The types implementing FromRaw, those that don't need further validation other than checking for the right structure, actually don't really need a "raw"-Version in a seperate module; they can be their own TryFromRaw::Raw types if they just implement Deserialize. 

This saves ~650 lines of code. Do you find this useful/interesting? Or is it a hard requirement for you that the FromRaw-types don't implement Deserialize? (The types that really do require additional validation still don't implement Deserialize)

Thanks for the fun exercise, in any case.
 
Summary of what this patch does:
- Instead of implementing FromRaw, implement Deserialize
- Add impl <T: DeserializeOwned>  TryFromRaw for T
  with Raw = T
- remove corresponding "raw"-modules from structs


lu-fennell:
Oh, sorry... I did not know about the "cargo check --all"; I just used "cargo check".

jplatte:
A previous version of my own branch had something like this, however I decided to keep the unconditional generation of raw types because then it's backwards compatible to add validation to an event type. If some event types implement Deserialize, you can also accidentally use them without `EventResult`, which would be unfortunate.

lu-fennell:
Ah I see... makes sense. Thanks for the explanation

jplatte:
Sure! Thanks for contributing anyway. Next time, I'd suggest discussing changes like this in a GitHub issue or on [#ruma:matrix.org](https://matrix.to/#/#ruma:matrix.org) before putting in the implementation work. (unless you're doing the work for exercise of course :slightly_smiling_face:)

PR #38, opened 2019-11-24
Add validation to NameEvent: name should not exceed 255 bytes

lu-fennell:


jplatte:
The CI failure is fixed now on master, it has nothing to do with your changes. However, there are a few things to improve still.

src/room/name.rs
```diff
@@ -227,14 +223,15 @@ pub(crate) mod raw {
 #[cfg(test)]
 mod tests {
     use std::convert::TryFrom;
+    use std::iter::FromIterator;
```
jplatte: You can nest these imports: `use std::{convert::TryFrom, iter::FromIterator};`

lu-fennell: Done

src/room/name.rs
```diff
@@ -251,6 +265,42 @@ mod tests {
...
     }
 
+    #[test]
+    fn name_fails_validation_when_too_long() {
+        // "XXXX .." 256 times
+        let long_string: String = String::from_iter(std::iter::repeat('X').take(256));
+        assert_eq!(long_string.len(), 256);
+
+        let long_content_json_string: String =
+            serde_json::json!({ "name": long_string.as_str()}).to_string();
```
jplatte:
```suggestion
            serde_json::json!({ "name": long_string.as_str() }).to_string();
```

src/room/name.rs
```diff
@@ -251,6 +265,42 @@ mod tests {
...
+        let from_raw: EventResult<NameEventContent> =
+            serde_json::from_str(long_content_json_string.as_str()).unwrap();
+
+        let result = from_raw.into_result();
+        assert!(result.is_err(), "Result should be invalid: {:?}", result);
+    }
+
+    #[test]
+    fn json_with_empty_name_creates_content_as_none() {
+        let long_content_json_string: String = serde_json::json!({ "name": ""}).to_string();
```
jplatte:
```suggestion
        let long_content_json_string: String = serde_json::json!({ "name": "" }).to_string();
```

src/room/name.rs
```diff
@@ -43,28 +43,41 @@ pub struct NameEventContent {
...
     type Raw = raw::NameEvent;
+    type Err = InvalidInput;
 
-    fn from_raw(raw: raw::NameEvent) -> Self {
-        Self {
-            content: FromRaw::from_raw(raw.content),
+    fn try_from_raw(raw: Self::Raw) -> Result<Self, Self::Err> {
+        let content = TryFromRaw::try_from_raw(raw.content)?;
+
+        let prev_content = match raw.prev_content {
```
jplatte: You can use [Result::transpose](https://doc.rust-lang.org/std/result/enum.Result.html#method.transpose) here :)

src/room/name.rs
lu-fennell: Done. And now also `&long_string`.

src/room/name.rs
lu-fennell: Done

src/room/name.rs
lu-fennell: Cool!

PR #49, opened 2020-02-28
room/encrypted: Fix the OlmV1 deserialization.

poljar:
A m.room.encrypted event with the algorithm type
"m.olm.v1.curve25519-aes-sha2" will have multiple ciphertext values.

The documentation for the struct field even mentions that the
ciphertext is a map, yet it only expected one body and one message type.

This patch fixes the de-serialization and puts all the available
ciphertext values into a hashmap.

### DIFF
--- a/src/room/encrypted.rs
+++ b/src/room/encrypted.rs
@@ -4,6 +4,7 @@ use js_int::UInt;
 use ruma_identifiers::{DeviceId, EventId, RoomId, UserId};
 use serde::{ser::SerializeStruct, Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::{from_value, Value};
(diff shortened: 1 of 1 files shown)


src/room/encrypted.rs
```diff
@@ -4,6 +4,7 @@ use js_int::UInt;
 use ruma_identifiers::{DeviceId, EventId, RoomId, UserId};
 use serde::{ser::SerializeStruct, Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::{from_value, Value};
+use std::collections::HashMap;
```
jplatte: Please move this into its own import group (separated by a blank line) above the other imports.

jplatte:
Thanks!

PR #51, opened 2020-03-08
Fill tests for ruma_events_macros

ldm0:
Part of #31 solving.

### DIFF
--- a/ruma-events-macros/tests/ruma_events_macros.rs
+++ b/ruma-events-macros/tests/ruma_events_macros.rs
@@ -1,11 +1,14 @@
 use std::{
     borrow::Cow,
-    convert::Infallible,
+    convert::{Infallible, TryFrom},
     fmt::{Debug, Display, Formatter, Result as FmtResult},
 };
 
+use js_int::UInt;
+use ruma_events_macros::ruma_event;
+use ruma_identifiers::{EventId, RoomAliasId, RoomId, UserId};
 use serde::{de::DeserializeOwned, Deserialize, Deserializer, Serialize};
-use serde_json::Value;
+use serde_json::{json, Value};
 
 /// The type of an event.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


ruma-events-macros/tests/ruma_events_macros.rs
```diff
@@ -233,173 +233,209 @@ impl Display for InvalidEvent {
 }
 
 // See note about wrapping macro expansion in a module from `src/lib.rs`
-pub mod common_case {
-    use std::convert::TryFrom;
-
+#[cfg(test)]
```
jplatte: `#[cfg(test)]` is redundant in files inside `tests`.

src/lib.rs
```diff
@@ -789,73 +789,73 @@ impl From<Algorithm> for String {
     }
 }
 
+#[rustfmt::skip]
```
jplatte: Please don't do this, when it is used on such a large scale, formatting will degrade over time.

src/lib.rs
```diff
@@ -789,73 +789,73 @@ impl From<Algorithm> for String {
...
 #[cfg(test)]
-mod tests {
+mod event_type_tests {
```
jplatte: Why did you rename the `tests` module?

ruma-events-macros/tests/ruma_events_macros.rs
ldm0: Oh sorry I didn't realize that.

src/lib.rs
ldm0: I think the 80 width is to small. and will make the test below less readable. I'll revert that.

src/lib.rs
ldm0: I think we need a more sepecific name of this test module. If name `tests` is used, We will get test result like this:
```
...
test tests::algorithms_serialize_and_deserialize_from_display_form ... ok
test tests::event_types_serialize_and_deserialize_from_display_form ... ok
...
```
Which may not that specific? (I guess?) If that's improper, I will revert that.

src/lib.rs
jplatte: The default max line width is 100, actually. We might change that if it makes the code more readable, but that shouldn't be part of this PR.

ldm0: Ok I get it. Thanks for reviewing! I have revert this and run the rustfmt.

src/lib.rs
jplatte: I think `mod tests` is a Rust convention, I can't remember a test module named something else. It could be a submodule of another module though. `lib.rs` is already quite big actually. How about this: You could create a new PR moving `EventType` and `Algorithm` to their own module respectively, and `pub use` them from the crate root. Then you can rebase this PR on top of that and avoid further increasing `lib.rs`' size while having clearer test names.

ldm0: Thanks for pointing that out! Working on it.

jplatte:
Now that the `serde_eq` macro is part of ruma-events, do you want to rebase and clean this up?

ldm0:
Yep! Working on it.

ldm0:
Since the serde_eq macro is in the ruma-event crate and not exported. I just left the macro used in this test as it is.

jplatte:
I think the test macros / functions could be public only for tests if the containg module is:

```rust
#[cfg(not(test)]
mod util;
#[cfg(test)]
pub mod util;
```

Apart from that, it would be nice if these tests would also use `json!` literals.

ldm0:
Ah. That would make things easier. Thanks a lot. Working on it.

jplatte:
@ldm0 actually, that might not work. You'll have to try. I don't know whether the test configuration is active on the main crate when compiling it for external tests.

ldm0:
Yep, that doesn't work.

ldm0:
I can't come up with a good function name. So I use: `serde_json_eq_try_from_raw`.

Issue #50, opened 2020-02-28
To-device events are missing a collection.


poljar:
To-device messages are sent using the [sendToDevice](https://matrix.org/docs/spec/client_server/latest#put-matrix-client-r0-sendtodevice-eventtype-txnid) API endpoint. Those events will end up in a sync response in the `to_device` field.

An event that is contained in the to_device field of the sync response, as per [spec](https://matrix.org/docs/spec/client_server/latest#extensions-to-sync) will have the following fields:

* sender
* type
* content

Many of the events that are sent mainly as to-device events are implemented in ruma-events without the `sender` field. Examples for this are:

* [Dummy](https://github.com/ruma/ruma-events/blob/master/src/dummy.rs#L20)
* [RoomKeyRequest](https://github.com/ruma/ruma-events/blob/master/src/room_key_request.rs#L12)
* [RoomKey](https://github.com/ruma/ruma-events/blob/master/src/room_key.rs)
* [ForwardedRoomKey](https://github.com/ruma/ruma-events/blob/master/src/forwarded_room_key.rs#L11)
* I believe all of the [key verification events](https://github.com/ruma/ruma-events/blob/master/src/key/verification/accept.rs#L13)

Another a bit of a different case, the [EncryptedEvent](https://github.com/ruma/ruma-events/blob/master/src/room/encrypted.rs#L40), which corresponds to a `m.room.encrypted` event is implemented as a `RoomEvent` containing a `sender`, `event_id`, `origin_server_ts` among others but the spec states [this](https://matrix.org/docs/spec/client_server/latest#m-room-encrypted):

> This event type is used when sending encrypted events. It can be used either within a room (in which case it will have all of the Room Event fields), or as a to-device event.

Historically the `m.olm.v1.curve25519-aes-sha2` algorithm was sent as a room event but since the introduction of the `m.megolm.v1.aes-sha2` algorithm, `m.olm.v1` events are sent as to-device events while `m.megolm.v1` events are sent as room events.

Note that key verification events can end up as to-device as well as room events as well, a separate issue exists for this in #33.

jplatte:
Fixed by #52.

PR #52, opened 2020-03-09
Implement a collection for to-device events.

poljar:
This implements the bulk part of #50 

Dummy events are prominently missing since the `Empty` struct doesn't seem to implement `FromRaw`, not sure if I'm just missing something there or if that needs to be fixed first.

Tests have been added for every implemented event type, which uncovered a small issue in the room key request deserialization, this is fixed as well.

### DIFF
--- a/src/to_device.rs
+++ b/src/to_device.rs
@@ -279,14 +279,14 @@ mod tests {
     #[test]
     fn room_key() {
         let room_key = r#"{
-        "content": {
-            "algorithm": "m.megolm.v1.aes-sha2",
-            "room_id": "!test:localhost",
-            "session_id": "fake_id",
-            "session_key": "fake_key"
-        },
-        "sender": "@alice:example.org",
-        "type": "m.room_key"
+            "content": {
+                "algorithm": "m.megolm.v1.aes-sha2",
+                "room_id": "!test:localhost",
+                "session_id": "fake_id",
+                "session_key": "fake_key"
+            },
+            "sender": "@alice:example.org",
+            "type": "m.room_key"
         }"#;
 
         let event = deserialize! {room_key, ToDevice::RoomKey};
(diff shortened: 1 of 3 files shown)


jplatte:
> Dummy events are prominently missing since the `Empty` struct doesn't seem to implement `FromRaw`, not sure if I'm just missing something there or if that needs to be fixed first.

I'm also not sure right now. I think you could have a variant on the raw event collection that just uses the same Dummy event as the non-raw event collection and simply skip the conversion function in all the matches. Would that work?

poljar:
I think that didn't work either but I forgot the exact error message. I won't be needing dummy events just yet so we can let this linger for a while if you're ok with it as well.

jplatte:
If you want to skip Dummy events, please add a comment that they should logically be present.

src/to_device.rs
```diff
@@ -0,0 +1,599 @@
...
+    fn room_key() {
+        let room_key = r#"{
+        "content": {
+            "algorithm": "m.megolm.v1.aes-sha2",
+            "room_id": "!test:localhost",
+            "session_id": "fake_id",
+            "session_key": "fake_key"
+        },
+        "sender": "@alice:example.org",
+        "type": "m.room_key"
```
jplatte:
```suggestion
            "content": {
                "algorithm": "m.megolm.v1.aes-sha2",
                "room_id": "!test:localhost",
                "session_id": "fake_id",
                "session_key": "fake_key"
            },
            "sender": "@alice:example.org",
            "type": "m.room_key"
```

src/to_device.rs
```diff
@@ -0,0 +1,599 @@
...
+        key::verification::{
+            cancel::CancelCode, start::StartEventContent, HashAlgorithm, KeyAgreementProtocol,
+            MessageAuthenticationCode, ShortAuthenticationString, VerificationMethod,
+        },
+        room::encrypted::EncryptedEventContent,
+        room_key_request::Action,
+        Algorithm, EventResult,
+    };
+
+    macro_rules! deserialize {
```
jplatte: Please re-format this macro so none of the lines exceed 100 characters.

jplatte:
:tada:

PR #53, opened 2020-03-09
Reorganize EventType and Algorithm.

ldm0:
Move `EventType` and `Algorithm` out of `src/lib.rs`.

PR #54, opened 2020-03-10
Add more tests for Algorithm and EventType.

ldm0:
Used to be a part of #53


jplatte:
You were talking about (`Partial`)`Eq` implementations being useful for the tests before, is that no longer applicable?

src/algorithm.rs
```diff
@@ -61,3 +61,36 @@ impl From<Algorithm> for String {
...
+#[cfg(test)]
+mod tests {
+    use serde_json::{from_str, to_string};
+
+    use super::*;
+
+    macro_rules! serde_eq {
+        ($type:ty, $de:literal, $se:expr) => {
+            assert_eq!($de, to_string(&$se).unwrap());
+            assert_eq!(from_str::<$type>($de).unwrap(), $se);
```
jplatte: You can drop the type macro parameter if you help type inference a bit:

```suggestion
            let mut val = $se;
            assert_eq!($de, to_string(&val).unwrap());
            val = from_str($de).unwrap();
            assert_eq!(val, $se);
```

ldm0:
Thanks for reviewing!

> You were talking about (`Partial`)`Eq` implementations being useful for the tests before, is that no longer applicable?

That is about the WIP tests in `ruma_events::{collections::{all, only}, stripped}`. Which is not relevant to this pull request.

src/algorithm.rs
ldm0: Ah! Sweet!

src/lib.rs
```diff
@@ -133,6 +133,9 @@ pub use self::{custom::CustomEvent, custom_room::CustomRoomEvent, custom_state::
 
 #[macro_use]
 mod macros;
```
jplatte: I didn't think about the existing module that would fit as a place for the new macro. Sorry for asking you to move it again, but could you move it to the macros module and put `#[cfg(test)]` on the macro itself?

src/test_util.rs
```diff
@@ -0,0 +1,8 @@
+macro_rules! serde_eq {
+    ($de:literal, $se:expr) => {
+        let mut val = $se;
+        assert_eq!($de, to_string(&val).unwrap());
```
jplatte:
```suggestion
        assert_eq!($de, serde_json::to_string(&val).unwrap());
```

src/test_util.rs
```diff
@@ -0,0 +1,8 @@
...
+        val = from_str($de).unwrap();
```
jplatte:
```suggestion
        val = serde_json::from_str($de).unwrap();
```

src/lib.rs
ldm0: Thanks for reviewing, working on it.

src/test_util.rs
ldm0: Thanks! Sorry for that.

src/algorithm.rs
```diff
@@ -61,3 +61,21 @@ impl From<Algorithm> for String {
         algorithm.to_string()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn algorithms_serialize_and_deserialize_from_display_form() {
```
jplatte:
```suggestion
    fn serialize_and_deserialize_from_display_form() {
```

Since this is now inside the `algorithm` module, the prefix can be removed.

src/event_type.rs
```diff
@@ -266,3 +266,81 @@ impl From<EventType> for String {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[allow(clippy::cognitive_complexity)]
+    #[test]
+    fn event_types_serialize_and_deserialize_from_display_form() {
```
jplatte:
```suggestion
    fn serialize_and_deserialize_from_display_form() {
```

jplatte:
@ldm0 Thanks! :)

PR #55, opened 2020-03-10
Implement helper method and more tests for MemberEvent.

### LINKED ISSUE #35: RoomMember events should implement a helper method to calculate the membership change.
`m.room.member` events indicate that the membership state of a user has changed.

The change is calculated by looking at the previous content of the event and comparing it to the current one. A [table](https://matrix.org/docs/spec/client_server/latest#m-room-member) in the Matrix spec lists all the valid options and what they mean.

It would be nice if `RoomMember` events would have a helper method that implements the logic in the mentioned table.

For this a new enum would be created that lists all the valid options (kick, profile_change, join, leave...) And the method would return an enum variant after it compares the previous and current content of the event.

ldm0:
Closes #35 
Part of #31 solving.

### DIFF
--- a/src/room/member.rs
+++ b/src/room/member.rs
@@ -168,6 +168,9 @@ pub enum MembershipChange {
     /// `displayname` or `avatar_url` changed.
     ProfileChanged,
 
+    /// Not implemented.
+    NotImplemented,
+
     /// Additional variants may be added in the future and will not be considered breaking changes
     /// to ruma-events.
     #[doc(hidden)]
@@ -208,7 +211,10 @@ impl MemberEvent {
             (Join, Ban) => MembershipChange::KickedAndBanned,
             (Leave, Invite) => MembershipChange::Invited,
             (Ban, Leave) => MembershipChange::Unbanned,
-            _ => MembershipChange::__Nonexhaustive,
+            (Knock, _) | (_, Knock) => MembershipChange::NotImplemented,
+            (__Nonexhaustive, _) | (_, __Nonexhaustive) => {
+                panic!("__Nonexhaustive enum variant is not intended for use.")
+            }
         }
     }
 }
--- a/src/room/member.rs
+++ b/src/room/member.rs
@@ -128,3 +128,305 @@ pub struct SignedContent {
     /// The token property of the containing third_party_invite object.
     pub token: String,
 }
+
+/// Translation of the membership change in `m.room.member` event.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
+pub enum MembershipChange {
+    /// No change.
+    None,
+
+    /// Must never happen.
+    Error,
+
+    /// User joined the room.
+    Joined,
+
+    /// User left the room.
+    Left,
+
+    /// User was banned.
+    Banned,
+
+    /// User was unbanned.
+    Unbanned,
+
+    /// User was kicked.
+    Kicked,
+
+    /// User was invited.
+    Invited,
+
+    /// User was kicked and banned.
+    KickedAndBanned,
+
+    /// User rejected the invite.
+    InvitationRejected,
+
+    /// User had their invite revoked.
+    InvitationRevoked,
+
+    /// `displayname` or `avatar_url` changed.
+    ProfileChanged,
+
+    /// Additional variants may be added in the future and will not be considered breaking changes
+    /// to ruma-events.
+    #[doc(hidden)]
+    __Nonexhaustive,
+}
+
+impl MemberEvent {
+    /// Helper function for memebership change. Check [the specification][spec] for details.
+    ///
+    /// [spec]: https://matrix.org/docs/spec/client_server/latest#m-room-member
+    pub fn membership_change(&self) -> MembershipChange {
+        use MembershipState::*;
+        let prev_membership = if let Some(prev_content) = &self.prev_content {
+            prev_content.membership
+        } else {
+            Leave
+        };
+        match (prev_membership, &self.content.membership) {
+            (Invite, Invite) | (Leave, Leave) | (Ban, Ban) => MembershipChange::None,
+            (Invite, Join) | (Leave, Join) => MembershipChange::Joined,
+            (Invite, Leave) => {
+                if self.sender.to_string() == self.state_key {
+                    MembershipChange::InvitationRevoked
+                } else {
+                    MembershipChange::InvitationRejected
+                }
+            }
+            (Invite, Ban) | (Leave, Ban) => MembershipChange::Banned,
+            (Join, Invite) | (Ban, Invite) | (Ban, Join) => MembershipChange::Error,
+            (Join, Join) => MembershipChange::ProfileChanged,
+            (Join, Leave) => {
+                if self.sender.to_string() == self.state_key {
+                    MembershipChange::Left
+                } else {
+                    MembershipChange::Kicked
+                }
+            }
+            (Join, Ban) => MembershipChange::KickedAndBanned,
+            (Leave, Invite) => MembershipChange::Invited,
+            (Ban, Leave) => MembershipChange::Unbanned,
+            _ => MembershipChange::__Nonexhaustive,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::convert::TryFrom;
+
+    use js_int::UInt;
+    use ruma_identifiers::{EventId, RoomId, UserId};
+    use serde_json::json;
+
+    use super::*;
+    use crate::util::serde_json_eq_try_from_raw;
+
+    #[test]
+    fn serde_with_no_prev_content() {
+        let event = MemberEvent {
+            content: MemberEventContent {
+                avatar_url: None,
+                displayname: None,
+                is_direct: None,
+                membership: MembershipState::Join,
+                third_party_invite: None,
+            },
+            event_id: EventId::try_from("$h29iv0s8:example.com").unwrap(),
+            origin_server_ts: UInt::new(1).unwrap(),
+            room_id: Some(RoomId::try_from("!n8f893n9:example.com").unwrap()),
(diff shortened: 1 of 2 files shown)


ldm0:
Currently this is blocked on merging of #54 . I will move the macro for testing into test_utils module after it.

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+
+#[cfg(test)]
+mod tests {
+    use js_int::UInt;
+    use ruma_identifiers::{EventId, RoomId, UserId};
+    use std::convert::TryFrom;
+
+    use crate::EventResult;
+
+    use super::*;
```
jplatte:
```suggestion
    use crate::EventResult;
    use super::*;
```

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+            MembershipChange::None
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use js_int::UInt;
+    use ruma_identifiers::{EventId, RoomId, UserId};
+    use std::convert::TryFrom;
```
jplatte: Please always move `std` import(s) to their own (empty-line separated) import group.

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+    /// Additional variants may be added in the future and will not be considered breaking changes
+    /// to ruma-events.
+    #[doc(hidden)]
+    __Nonexhaustive,
+}
+
+impl MemberEvent {
+    /// Helper function for memebership change. Check [spec] for details
+    ///
+    /// [spec]:https://matrix.org/docs/spec/client_server/latest#m-room-member
```
jplatte:
```suggestion
    /// Helper function for memebership change. Check [the specification][spec] for details.
    ///
    /// [spec]: https://matrix.org/docs/spec/client_server/latest#m-room-member
```

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+                    if self.sender.to_string() == self.state_key {
+                        MembershipChange::Left
+                    } else {
+                        MembershipChange::Kicked
+                    }
+                }
+                (Join, Ban) => MembershipChange::KickedAndBanned,
+                (Leave, Invite) => MembershipChange::Invited,
+                (Ban, Leave) => MembershipChange::Unbanned,
+                _ => MembershipChange::__Nonexhaustive,
```
jplatte: Why should there be a catch-all branch here?

jplatte:
@poljar Do you want to review this?

poljar:
Sure, I'll take a look tomorrow.

jplatte:
Thanks!

src/room/member.rs
ldm0: Thanks for reviewing.

src/room/member.rs
ldm0: Some membership change in spec is `Not implemented.`. So I'd like to catch them all and return the __Nonexhaustive.

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+            }
+        } else {
+            MembershipChange::None
```
poljar: This isn't quite right, or as it is the method isn't as useful as it could be.

If there is no previous content but the state in the current content is a `Join` then the change `Joined`

Ideally client authors would never need to look up the individual states in the content and just rely on this method.

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+    ///
+    /// [spec]: https://matrix.org/docs/spec/client_server/latest#m-room-member
+    pub fn membership_change(&self) -> MembershipChange {
+        use MembershipState::*;
+        if let Some(prev_content) = &self.prev_content {
+            match (prev_content.membership, &self.content.membership) {
+                (Invite, Invite) | (Leave, Leave) | (Ban, Ban) => MembershipChange::None,
+                (Invite, Join) | (Leave, Join) => MembershipChange::Joined,
+                (Invite, Leave) => {
+                    if self.sender.to_string() == self.state_key {
```
poljar: I know this isn't your fault and it doesn't need to be fixed in this PR, but it's quite annoying that we can't get a `&str` out of a `UserId`. This will unnecessarily allocate stuff.

src/room/member.rs
```diff
@@ -128,3 +128,213 @@ pub struct SignedContent {
...
+            },
+            event_id: EventId::try_from("$h29iv0s8:example.com").unwrap(),
+            origin_server_ts: UInt::new(1).unwrap(),
+            room_id: Some(RoomId::try_from("!n8f893n9:example.com").unwrap()),
+            sender: UserId::try_from("@carl:example.com").unwrap(),
+            state_key: "example.com".to_string(),
+            unsigned: None,
+            prev_content: None,
+        };
+        let expect = r#"{"content":{"membership":"join"},"event_id":"$h29iv0s8:example.com","origin_server_ts":1,"room_id":"!n8f893n9:example.com","sender":"@carl:example.com","state_key":"example.com","type":"m.room.member"}"#;
```
poljar: I know that this isn't particularly important but, can we get this formatted a bit better? It's a bit hard to see what's going on.

There's no reason to trim all the white-space. I did a similar thing [here](https://github.com/ruma/ruma-events/pull/52/files#diff-030ed025a789cd63963735cbb156b899R604).

src/room/member.rs
ldm0: Thanks for reviewing. Working on it. But the problem is this is not in the spec.

poljar: It's mentioned in the sentence before the table:
> Previous membership can be retrieved from the prev_content object on an event. If not present, the user's previous membership must be assumed as leave.

ldm0: Ah! I didn't noticed that. Thanks for pointing that out.

src/room/member.rs
ldm0: Ah, that really a problem. I actually have tried to solve it but with not luck. If only check deserialization, we can use the formatted json. But if we want to test serialization, the order of k-v pair is really important (from my point of view).
Method 1. Deserialize the formatted json and compare with expected Value. The serialization result cannot be tested.
Method 2. Deserialize the formatted json and then serialize it to unformatted json, then trying compare it with the serlization result of expect value. We will notice that the order of k-v pair of the json we deserialize and then serialize sometimes changed. That's because the serde_json::Value is using BTreeMap and all fields is sorted alphabetically.
So I decided to leave it unformatted. :-/

poljar: But adding white-space wouldn't mess with the order, no? And if we test the serialization we could just remove all the white-space before we compare the strings?

ldm0: If there are white-space in the json string? Simply remove white-space would mess it. :-/

ldm0: I mean string literal in json.

ldm0: Ah I think I can write a function that skip string literal and remove ws. That would help. I will file it in another PR.

jplatte: You can rewrite the tests to use `json!()` literals. I just did this for the tests you added, also rewriting the macro `serde_eq!` to `fn serde_json_eq`. I think the same thing would be possible here, and improve readability as `json!()` literals can easily be split across multiple lines.

ldm0: But the order of json field cannot be tested after using `json!()`. If that's not important I think its ok.

ldm0: Using a function like this can as a pre process for string literal can also solve the problem, while still preserve test for the order of k-v pair:
```rust

/// Used to unformat json
#[cfg(test)]
pub fn shrink_json(json: &str) -> String {
    let json = json.as_bytes();
    let mut result = Vec::with_capacity(json.len());
    let mut begin = 0;
    // Out of string literal
    let mut out = true;
    for (i, ch) in json.iter().enumerate() {
        match *ch {
            b' ' | 0x9..=0xd => {
                if out {
                    result.extend(&json[begin..i]);
                    begin = i + 1;
                }
            }
            b'"' => out = !out,
            _ => {}
        }
    }
    result.extend(&json[begin..]);
    unsafe { String::from_utf8_unchecked(result) }
}
```

jplatte: It's not only not important, the tests should succeed independently of the serialized order of the fields.

ldm0: Without specifing order of the field, same event can be serialized to json with different order. My only concern is some program may depend on the order of event json field.

ldm0: That's fine. Working on it.

jplatte: That would be extremely weird for anything working with JSON. Also we can't really stop someone up the dependency tree from activating `serde_json`s `preserve_order` feature and thus changing our field order.

src/room/member.rs
jplatte: Do we have an issue for that on ruma-identifiers?

src/room/member.rs
jplatte: Okay, well then I still would want this branch to be `(Knock, _) | (_, Knock) =>` rather than `_ =>` to make sure no possible case is accidentally forgotten.

But also, I don't like that this hidden value uses the "non-exhaustive" terminology. If we couldn't convert the code to `#[non_exhaustive]` and drop this variant in the process, I don't think the variant should be named that. I think we should have a `NotImplemented` variant that is returned here in addition to `__Nonexhaustive`, but I'm not certain whether it should be public or de-facto private (via `#[doc(hidden)]`). @poljar what do you think?

src/room/member.rs
poljar: No, I wanted to come up with a sensible proposal before opening one.

src/room/member.rs
poljar: I guess making it public won't hurt that much since the spec could always get more states. That being said I don't have strong feelings about this besides that every known case should be explicitly handled.

PR #58, opened 2020-03-16
Update unsigned field's type from `Option<Value>` to `Map<String, Value>`

jplatte:
* it's guaranteed to always be an object
* `Option` isn't needed because the distinction empty object / missing field is not of interest


jplatte:
Status: This has to be rebased after `Serialize` implementations are no longer generated by us but by `serde`s own derive macros.

PR #63, opened 2020-03-21
Dummy event for to-device collection

### LINKED ISSUE #56: Dummy event type missing from to-device collection
See the comments in #52.

DevinR528:
Resolves #56


jplatte:
I wrote in #52:

> I think you could have a variant on the raw event collection that just uses the same Dummy event as the non-raw event collection and simply skip the conversion function in all the matches.

Did you try that? Was there a problem with it?

DevinR528:
Like this commit? In the `raw` mod I have to use the `from_value` func to get the `ToDeviceEvent.sender` info out of the json body right?

jplatte:
I was commenting on your use of `Empty` but now I see it's the same as `DummyEventContent`. Still it would be clearer to use the type alias.

With your latest change you did implement exactly what I was thinking regarding conversions, but since the previous version also worked and was consistent with the other branches, I don't see why we would actually skip the conversion (even though it's currently a no-op).

DevinR528:
The conversion functions are really cool I didn't know you could do that with enums thats a nifty trick.

src/to_device.rs
```diff
@@ -289,6 +301,19 @@ mod tests {
...
+    fn dummy() {
+        let dummy = r#"{
+            "content": {},
+            "sender": "@alice:example.org",
+            "type": "m.dummy"
+        }"#;
+
+        let event = deserialize! {dummy, AnyToDeviceEvent::Dummy};
+
+        assert_eq!(event.content, Empty,);
```
jplatte:
```suggestion
        assert_eq!(event.content, DummyEventContent);
```

jplatte:
> The conversion functions are really cool I didn't know you could do that with enums thats a nifty trick.

Yeah, and it [works for tuple structs too](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=282b6f009cf3508bc976a5265dfcf5639) :)

src/to_device.rs
DevinR528: hmm compiler says `can't use type alias as constructor`

jplatte: Oh, interesting... Revert that then, but without the trailing `,`.

Issue #61, opened 2020-03-19
Declare custom events without the ruma_event! macro


jplatte:
Custom events currently require some special-casing to work with `ruma_event!`. We sould get rid of this special-casing and instead declare them manually.

PR #64, opened 2020-03-23
derive Serialize in `ruma_events!`, use json EventType string for `event_type` field

### LINKED ISSUE #62: Don't generate `Serialize` implementations manually
This requires somewhat extensive changes to `ruma_event!` in that the `event_type` needs to become a string literal (the serialized string form of `EventType`) so we can use the (undocumented... https://github.com/serde-rs/serde-rs.github.io/pull/103) attribute `#[serde(tag)]` on the generated event structs. With that change, custom events would require even more special-casing, so I think they should instead be declared without the use of `ruma_event!` (#61).

Update: after #65 and #64, all that remains to be done is removing the `Serialize` implementations from our generated code, instead deriving them using the appropriate attribute(s) mentioned above; and add tests.

DevinR528:
Partially resolves #62 

Don't bother looking at it quite yet I'm still working. I'll start up again in a couple hours.


ruma-events-macros/src/gen.rs
```diff
@@ -90,15 +90,23 @@ impl ToTokens for RumaEvent {
...
-            let event_type = &self.event_type;
-
+            let event_type = self
+                .event_type
+                .value()
+                .split('.')
+                .filter(|s| *s != "m")
+                .collect::<Vec<_>>()
+                .join("_")
+                .to_camel_case();
```
jplatte: Does this work too?

```suggestion
            assert_eq!(self.event_type[2..], "m.");
            let event_type = self.event_type.value().to_camel_case();
```

ruma-events-macros/src/gen.rs
```diff
@@ -90,15 +90,23 @@ impl ToTokens for RumaEvent {
...
+
+            let variant = syn::parse_str::<Ident>(&event_type).unwrap();
```
jplatte: For something as simple as constructing an `Ident`, I'd prefer to use its constructor rather than syn's parsing functionality.

ruma-events-macros/src/gen.rs
```diff
@@ -450,8 +458,8 @@ fn populate_state_fields(is_custom: bool, content_name: Ident, fields: Vec<Field
 }
 
 /// Checks if the given `Path` refers to `EventType::Custom`.
-fn is_custom_event_type(event_type: &Path) -> bool {
-    event_type.segments.last().unwrap().ident == "Custom"
+fn is_custom_event_type(event_type: &LitStr) -> bool {
+    event_type.value().contains("custom")
```
jplatte: Custom events are ones without a known event type. Updating them to some fake event type containing "custom" would be misleading. I didn't think about this when suggesting you could work on #62. This code thus has to be removed before this can be merged (#61). Do you mind trying that in another PR?

DevinR528: I'll start working on it in a separate PR.

ruma-events-macros/src/gen.rs
DevinR528: It worked much more readable.

ruma-events-macros/src/gen.rs
```diff
@@ -90,15 +90,16 @@ impl ToTokens for RumaEvent {
...
 
-        let event_type = if self.is_custom {
+        let event_type_variant = if self.is_custom {
             quote! {
                 ::ruma_events::EventType::Custom(self.event_type.clone())
             }
         } else {
-            let event_type = &self.event_type;
-
+            assert_eq!(&self.event_type.value()[..2], "m.");
```
jplatte: It would be good to only evaluate `self.event_type.value()` once as it allocates a new string for each call.

ruma-events-macros/src/gen.rs
```diff
@@ -411,6 +407,16 @@ fn is_option(ty: &Type) -> bool {
     }
 }
 
+/// Splits the given `event_type` string on `.` and `_` removing the `m.` then
+/// camel casing to give the `EventType` variant.
+fn to_camel_case(name: String) -> String {
+    assert_eq!(&name[..2], "m.");
+    name[2..]
+        .split(&['.', '_'] as &[char])
+        .map(|s| s.chars().next().unwrap().to_uppercase().collect::<String>() + &s[1..])
```
jplatte:
```suggestion
        .map(|s| s.chars().next().unwrap().to_uppercase().to_string() + &s[1..])
```

PR #65, opened 2020-03-23
remove special handling of custom event type

### LINKED ISSUE #61: Declare custom events without the ruma_event! macro

DevinR528:
Resolves #61 
Partially relates to #62


src/lib.rs
```diff
@@ -153,6 +151,12 @@ pub mod collections {
         pub mod only;
     }
 }
+///
+pub mod custom {
+    pub mod basic;
+    pub mod room;
+    pub mod state;
+}
```
DevinR528: Is the mod structure okay here?

src/lib.rs
```diff
@@ -406,53 +411,53 @@ pub trait StateEvent: RoomEvent {
...
+//     use ruma_events_macros::ruma_event;
+//     use serde_json::Value;
+
+//     ruma_event! {
+//         /// A custom state event not covered by the Matrix specification.
+//         CustomStateEvent {
+//             kind: StateEvent,
+//             event_type: Custom,
+//             content_type_alias: {
+//                 /// The payload for `CustomStateEvent`.
```
DevinR528: TODO remove commented out code

ruma-events-macros/src/gen.rs
```diff
@@ -181,9 +166,7 @@ impl ToTokens for RumaEvent {
             serialize_field_calls.push(serialize_field_call);
         }
 
-        let (manually_serialize_type_field, import_event_in_serialize_impl) = if self.is_custom {
-            (TokenStream::new(), TokenStream::new())
-        } else {
+        let (manually_serialize_type_field, import_event_in_serialize_impl) = {
```
jplatte: There's no longer a reason to initialize these two together from a tuple.

src/custom/basic.rs
```diff
@@ -0,0 +1,67 @@
...
+    type Raw = raw::CustomEvent;
+    fn from_raw(raw: raw::CustomEvent) -> Self {
+        Self {
+            content: FromRaw::from_raw(raw.content),
+            event_type: raw.event_type,
+        }
+    }
+}
+
+impl Serialize for CustomEvent {
```
jplatte: This impl can be derived by serde if the `event_type` field has `#[rename("type")]`.

src/custom/basic.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// The event's content.
+    pub content: CustomEventContent,
+    /// The custom type of the event.
+    pub event_type: String,
+}
+
+/// The payload for `CustomEvent`.
+pub type CustomEventContent = Value;
+
+impl FromRaw for CustomEvent {
```
jplatte: This impl can be derived, that's part of the reason I split `FromRaw` deriving from `ruma_event!`.

src/custom/basic.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// The event's content.
+    fn content(&self) -> &Self::Content {
+        &self.content
+    }
+    /// The type of the event.
+    fn event_type(&self) -> EventType {
+        EventType::Custom(self.event_type.clone())
+    }
+}
+/// "Raw" versions of the event and its content which implement `serde::Deserialize`.
```
jplatte:
```suggestion
}

/// "Raw" versions of the event and its content which implement `serde::Deserialize`.
```

src/custom/basic.rs
```diff
@@ -0,0 +1,67 @@
...
+pub(crate) mod raw {
+    use super::*;
+    /// A custom basic event not covered by the Matrix specification.
```
jplatte:
```suggestion
    use super::*;

    /// A custom event not covered by the Matrix specification.
```

src/custom/room.rs
```diff
@@ -0,0 +1,136 @@
+//! A custom room event outside of the Matrix specification.
+
+use crate::{Event, EventType, FromRaw, RoomEvent};
+use serde::{Deserialize, Serialize, Serializer};
+use serde_json::Value;
+/// A custom room event not covered by the Matrix specification.
```
jplatte:
```suggestion
use serde_json::Value;

/// A custom room event not covered by the Matrix specification.
```

src/lib.rs
jplatte: Why make a breaking change instead of keeping the current modules?

tests/ruma_events_macros.rs
```diff
@@ -121,18 +121,7 @@ mod common_case {
 
 mod custom_event_type {
```
jplatte: This test module can be removed completely as it was testing the codegen for custom events.

src/lib.rs
DevinR528: I thought there would be more code in each section but with all the derives its pretty short. will fix

src/lib.rs
```diff
@@ -406,53 +406,265 @@ pub trait StateEvent: RoomEvent {
...
+        /// A key that determines which piece of room state the event represents.
+        pub state_key: String,
+        /// Additional key-value pairs not signed by the homeserver.
+        pub unsigned: Option<serde_json::Value>,
+    }
+
+    /// The payload for `CustomStateEvent`.
+    pub type CustomStateEventContent = Value;
+
+    impl StateEvent for CustomStateEvent {
```
jplatte: Could you move this impl after the `RoomEvent` one?

src/lib.rs
```diff
@@ -406,53 +406,265 @@ pub trait StateEvent: RoomEvent {
...
+        /// The unique identifier for the user who sent this event.
+        pub sender: ruma_identifiers::UserId,
+        /// Additional key-value pairs not signed by the homeserver.
+        pub unsigned: Option<serde_json::Value>,
+    }
+
+    /// The payload for `CustomRoomEvent`.
+    pub type CustomRoomEventContent = Value;
+
+    impl RoomEvent for CustomRoomEvent {
```
jplatte: Could you move this impl after the `Event` one?

jplatte:
Thanks! Since I thought the patch I sent to you was part of this PR I added `Co-authored-by` in the squash commit (but the patch is part of #64, not this PR). I have reverted that commit and pushed another one without `Co-authored-by`.

DevinR528:
Thanks, in the future you don't have to worry about it if it's more work.

jplatte:
It would have been trivial, would I be able to disable branch protections. But it still wasn't much work, and I was just fixing what I broke, plus I consider authorship information important.

PR #68, opened 2020-03-29
Remove manual Serialize impl, use derive macro instead, Fix #62

### LINKED ISSUE #62: Don't generate `Serialize` implementations manually

NerdyPepper:
- Don't strip serde attrs on room event and state fields
 - Manually call `skip_serializing_if = Option::is_none` attr on `Option` type room event and room state fields
 - Modify tests to include the `type` field before all other fields

### DIFF
--- a/ruma-events-macros/src/gen.rs
+++ b/ruma-events-macros/src/gen.rs
@@ -217,7 +217,7 @@ impl ToTokens for RumaEvent {
         let event_type_name = self.event_type.value();
         let output = quote!(
             #(#attrs)*
-            #[derive(serde::Serialize, Clone, PartialEq, Debug, ruma_events_macros::FromRaw)]
+            #[derive(Clone, PartialEq, Debug, serde::Serialize, ruma_events_macros::FromRaw)]
             #[serde(rename = #event_type_name, tag = "type")]
             pub struct #name {
                 #(#event_fields),*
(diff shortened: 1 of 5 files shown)


ruma-events-macros/src/gen.rs
```diff
@@ -269,42 +214,19 @@ impl ToTokens for RumaEvent {
...
 
-        let stripped_fields = event_fields
-            .iter()
-            .map(|event_field| strip_serde_attrs(event_field));
-
+        let event_type_name = self.event_type.value();
         let output = quote!(
             #(#attrs)*
-            #[derive(Clone, PartialEq, Debug, ruma_events_macros::FromRaw)]
+            #[derive(serde::Serialize, Clone, PartialEq, Debug, ruma_events_macros::FromRaw)]
```
jplatte: I'd like to have the builtin derives first.

```suggestion
            #[derive(Clone, PartialEq, Debug, serde::Serialize, ruma_events_macros::FromRaw)]
```

jplatte:
Thanks!

PR #70, opened 2020-03-29
Rewrite tests to use `serde_json::json!` instead of raw strings

NerdyPepper:
- [x] event_type.rs
- [x] presence.rs
- [x] ignored_user_list.rs
- [x] push_rules.rs
- [x] to_device.rs
- [x] direct.rs
- [x] dummy.rs
- [x] stripped.rs
- [x] key/verification/start.rs
- [x] key/verification/cancel.rs
- [x] room/pinned_events.rs
- [x] room/name.rs
- [x] room/canonical_alias.rs
- [x] room/member.rs
- [x] room/create.rs
- [x] room/power_levels.rs
- [x] room/message.rs
- [x] room/encrypted.rs
- [x] room/server_acl.rs

### DIFF
--- a/src/direct.rs
+++ b/src/direct.rs
@@ -25,7 +25,9 @@ mod tests {
     use std::collections::HashMap;
 
     use ruma_identifiers::{RoomId, UserId};
-    use serde_json::{from_value, json, to_value};
+    use serde_json::from_value as from_json_value;
+    use serde_json::json;
+    use serde_json::to_value as to_json_value;
 
     use super::{DirectEvent, DirectEventContent};
     use crate::EventResult;
@@ -46,7 +48,7 @@ mod tests {
             "type": "m.direct"
         });
 
-        assert_eq!(to_value(&event).unwrap(), json_data);
+        assert_eq!(to_json_value(&event).unwrap(), json_data);
     }
 
     #[test]
@@ -64,7 +66,7 @@ mod tests {
             "type": "m.direct"
         });
 
-        let event: DirectEvent = from_value::<EventResult<_>>(json_data)
+        let event: DirectEvent = from_json_value::<EventResult<_>>(json_data)
             .unwrap()
             .into_result()
             .unwrap();
--- a/src/direct.rs
+++ b/src/direct.rs
@@ -25,9 +25,7 @@ mod tests {
     use std::collections::HashMap;
 
     use ruma_identifiers::{RoomId, UserId};
-    use serde_json::from_value as from_json_value;
-    use serde_json::json;
-    use serde_json::to_value as to_json_value;
+    use serde_json::{from_value as from_json_value, json, to_value as to_json_value};
 
     use super::{DirectEvent, DirectEventContent};
     use crate::EventResult;
--- a/src/direct.rs
+++ b/src/direct.rs
@@ -25,7 +25,7 @@ mod tests {
     use std::collections::HashMap;
 
     use ruma_identifiers::{RoomId, UserId};
-    use serde_json::to_string;
+    use serde_json::{from_value, json, to_value};
 
     use super::{DirectEvent, DirectEventContent};
     use crate::EventResult;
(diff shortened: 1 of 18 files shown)


src/direct.rs
```diff
@@ -25,7 +25,7 @@ mod tests {
     use std::collections::HashMap;
 
     use ruma_identifiers::{RoomId, UserId};
-    use serde_json::to_string;
+    use serde_json::{from_value, json, to_value};
```
jplatte: Please only import `json!` and use the other items by their absolute path. Alternatively, we could establish the convention of renaming them to `from_json_value` and `to_json_value` locally. What do you think?

src/push_rules.rs
```diff
@@ -444,37 +444,47 @@ mod tests {
...
 
     #[test]
     fn deserialize_tweak_sound_action() {
+        let json_data = json!({
+            "set_tweak": "sound",
+            "value": "default"
+        });
         assert_eq!(
-            from_str::<Action>(r#"{"set_tweak":"sound","value":"default"}"#).unwrap(),
+            from_value::<Action>(json_data).unwrap(),
```
jplatte: Feel free to remove the redundant generic arguments in this PR:

```suggestion
            from_value(json_data).unwrap(),
```

src/direct.rs
NerdyPepper: I am leaning towards the convention of renaming the imports. It would look a lot cleaner imo.

jplatte: Okay. Let's do that, then.

src/push_rules.rs
NerdyPepper: ah it seems we need that type parameter
 ```
error[E0283]: type annotations needed
   --> src/push_rules.rs:489:13
    |
489 |             from_json_value(json_data).unwrap(),
    |             ^^^^^^^^^^^^^^^--------------------
    |             |
    |             this method call resolves to `T`
    |             cannot infer type for type parameter `T` declared on the function `from_value`
    |             help: consider specifying the type argument in the function call: `from_json_value::<T>`
```

Comment (jplatte) [reply src/push_rules.rs]: Ah, I thought it could be inferred here. Nevermind then.

File: src/direct.rs [hunk 3]
```diff
@@ -25,7 +25,9 @@ mod tests {
     use std::collections::HashMap;
 
     use ruma_identifiers::{RoomId, UserId};
-    use serde_json::{from_value, json, to_value};
+    use serde_json::from_value as from_json_value;
+    use serde_json::json;
+    use serde_json::to_value as to_json_value;
```
jplatte: You don't need to split up the import to rename items in it.

```suggestion
    use serde_json::{from_value as from_json_value, json, to_value as to_json_value};
```

src/to_device.rs
```diff
@@ -627,23 +629,26 @@ mod tests {
...
+        });
 
         let event = deserialize! {source, AnyToDeviceEvent::KeyVerificationRequest};
         assert_eq!(event.content.transaction_id, "S0meUniqueAndOpaqueString");
         assert_eq!(event.content.from_device, "AliceDevice2");
         assert_eq!(event.content.methods, &[VerificationMethod::MSasV1]);
-        assert_eq!(event.content.timestamp, UInt::new(1559_598944869).unwrap());
+        assert_eq!(
+            event.content.timestamp,
+            UInt::new(1_559_598_944_869).unwrap()
```
jplatte: Thanks for fixing that ugly literal :)

NerdyPepper: :)

src/direct.rs
NerdyPepper: oh i see, ill fix that up real quick

src/key/verification/start.rs
```diff
@@ -415,10 +415,20 @@ mod tests {
             content: key_verification_start_content,
         };
 
-        assert_eq!(
-            to_string(&key_verification_start).unwrap(),
-            r#"{"content":{"from_device":"123","transaction_id":"456","method":"m.sas.v1","key_agreement_protocols":["curve25519"],"hashes":["sha256"],"message_authentication_codes":["hkdf-hmac-sha256"],"short_authentication_string":["decimal"]},"type":"m.key.verification.start"}"#
-        );
+        let json_data = json!({
+            "content":{
```
jplatte: Please add spaces after the `"key":` part here and below.

```suggestion
            "content": {
```

src/lib.rs
```diff
@@ -111,6 +111,7 @@
 
 #![warn(rust_2018_idioms)]
 #![deny(missing_debug_implementations, missing_docs)]
+#![recursion_limit = "1024"]
```
jplatte: Please move this two lines up, so it's not mixed in between the lint attributes.

src/key/verification/start.rs
NerdyPepper: alright, i have made these changes, i hope i got everything.

jplatte: If you haven't, that can still be fixed later.

Issue #69, opened 2020-03-29
Replace all remaining manual `Serialize` implementations by derives


jplatte:
#68 did the bulk of this, replacing `Serialize` implementations in `ruma_event!`. There's still a few events declared without `ruma_event!` that implement `Serialize` manually though. These should also be updated to use `#[derive(Deserialize)]` + `#[serde(rename = "…", tag = "type")]`.

jplatte:
Done in #72

PR #72, opened 2020-04-05
Replace all remaining manual  implementations by derives

PipsqueakH:
#69


jplatte:
Thanks! `IgnoredUserListEvent` can now be implemented with `ruma_event!` it seems – AFAICT, it was only defined the way it is because the macro didn't previously support serde attributes on content fields. If you want to do that now, just tell me, otherwise I'll merge this as-is once CI is happy, and that work can be done in a future PR.

PipsqueakH:
maybe a future PR is a better choice.

PR #75, opened 2020-04-18
Replace EventResult with OwningRef<JsonValue, _>

### LINKED ISSUE #71: Provide access for the underlying json for events
This is an important feature for building a homeserver on top of ruma-events.

For the incoming bits, we could "just" build on top of `EventResult`, making it store the underlying JSON *always* instead of just on error. However, that wouldn't solve the issue of relaying the full JSON of the event. @emilazy suggested in [#ruma:matrix.org](https://matrix.to/#/#ruma:matrix.org) some time ago that we could make our deserialized types "views into" the original JSON string and I think something like that could be applicable here. However, I'd still like the event types to have `'static` lifetime, which would mean storing the views alongside the JSON bits they reference – this also means we either have to resort to `unsafe` or add a crate like [rental](https://crates.io/crates/rental) to our dependencies, but I think that's fine.

jplatte:
Fixes #71.

*Work in progress*


jplatte:
After looking at the owning_ref crate in more detail, it doesn't seem so nice anymore :(

Also, it will have to be `Box<serde_json::Value>` because `serde_json::Value` itself can't be `StableDeref`. But that's a small issue in comparison.

jplatte:
Having looked into `owning_ref` more deeply now, I think I have a good understanding of the kind of API I want. Unfortunately, AFAICT it is impossible to provide this API without generic associated types. I'm currently writing my own crate similar to `owning_ref` using GAT and have already had to work around one compiler bug, but am pretty confident that what I'm writing is exactly what we need here. The only question then is whether we can block this feature on GAT...

Issue #47, opened 2020-02-20
Add event content collections


jplatte:
This is needed to have strongly-typed `content` fields in `ruma-client-api` endpoints that send events as opposed to receiving them. In those cases, a full event can't yet be created.

It probably makes sense to first change the current event collections from hand-written to macro-generated because they're already quite boilerplate-heavy. Then, the same macro invocation can be extended to also generate event content collections.

jplatte:
This is what I imagine the macro invocation to look like:

```rust
collections! {
    non_room_events: [
        Direct,
        Dummy,
        // ...
    ],
    message_events: [
        CallAnswer,
        CallCandidates,
        CallHangup,
        CallInvite,
        RoomEncrypted,
        RoomMessage,
        // ...
    ],
    state_events: [
        RoomAliases,
        RoomAvatar,
        RoomCanonicalAlias,
        RoomCreate,
        // ...
    ],
}
```

jplatte:
This is done in the api22 branch.

PR #78, opened 2020-04-22
room-messages: Add the ability to send encrypted room messages.

poljar:


jplatte:
Please run `cargo fmt`.

jplatte:
Hey @poljar, I've reverted this change. I merged it because I didn't really understand the details and assumed you knew what you were doing, but now have looked into the details and this no longer seems correct to me (also only serialization was supported, not deserialization).

I believe the actual change required to send encrypted room messages is to update `r0::message::create_message_event` (and `r0::client_exchange::send_event_to_device`) in ruma-client-api.

poljar:
> Hey @poljar, I've reverted this change. I merged it because I didn't really understand the details and assumed you knew what you were doing, but now have looked into the details and this no longer seems correct to me 

Huh? I don't really want to claim to know what I'm doing since I certainly don't and sorry if this comes off as overly verbose but let's go through how events and sending them is implemented in Ruma.

Matrix events get sent to the server as a string that denotes a type and a JSON map/dict of the content.

So for example a event with `m.room.message` for the type and the following dict for the content
```
{
    body: "Hello world
}
```

This is true for sending normal room events as well as for to-device events. Now Ruma has structs for fully build events, those will contain additional server provided data e.g. the timestamp, the sender, and structs for the event content.

The struct for the fully build event encapsulates the content structs to create per event type specific fully build events.

The event content structs represent the data that needs to be sent out by the clients. If we take out example from above, the [TextMessageEventContent](https://docs.rs/ruma-events/0.21.0/ruma_events/room/message/struct.TextMessageEventContent.html) maps to the content we need to send.

Since we want to support the sending of multiple event types in our [rooms/send](https://matrix.org/docs/spec/client_server/latest#put-matrix-client-r0-rooms-roomid-send-eventtype-txnid) request builder an enum over different event contents was [used](https://docs.rs/ruma-client-api/0.7.2/ruma_client_api/r0/message/create_message_event/struct.Request.html#structfield.data).

Now one could argue that [MessageEventContent](https://docs.rs/ruma-events/0.21.0/ruma_events/room/message/enum.MessageEventContent.html) is only meant to represent content of `m.room.message` events. But that would mean that the [create_message_event](https://docs.rs/ruma-client-api/0.8.0-rc.5/ruma_client_api/r0/message/create_message_event/struct.Request.html) is named and defined a bit unwisely, since a message in that context can and needs to be more than just the `m.room.message` subtypes. Really any type of event can be sent over this endpoint, most prominently `m.room.encrypted` and the `m.key.verification.*` types.

So the enum that we're using in the `create_message_event` requests needs to contain the `m.room.encrypted` content.

> I believe the correct change to send encrypted room messages is to update `r0::message::create_message_event` (and `r0::client_exchange::send_event_to_device`) in ruma-client-api.

Could you clarify what you mean by that? Update to what? If `MessageEventContent` is indeed only for `m.room.message` subtypes, I don't think we have any enum that encapsulates different event contents. So a new one would need to be defined.

```rust
pub enum MessageEvent {
    Message(MessageEventContent),
    Encrypted(EncryptedEventContent),
    KeyVerification(...),
}
```
It's unclear if or why this should be in ruma-client-api.

Could we also stick to namings from the spec? I think the whole confusion stems from the request builder being called `create_message_event` where message means more than `m.room.message` and using an enum called `MessageEventContent` for the data field where message means only `m.room.message`.

jplatte:
> Now one could argue that MessageEventContent is only meant to represent content of m.room.message events.

That's exactly what that enum is for.

> But that would mean that the [create_message_event](https://docs.rs/ruma-client-api/0.8.0-rc.5/ruma_client_api/r0/message/create_message_event/struct.Request.html) is named and defined a bit unwisely, since a message in that context can and needs to be more than just the `m.room.message` subtypes. Really any type of event can be sent over this endpoint, most prominently `m.room.encrypted` and the `m.key.verification.*` types.

Yes, that endpoint will have to be renamed, in addition to having its `data` request field updated to a different type. I think the reason for the current naming and typing is the unclear wording in the specification. It talks about message events when really, AFAICT, it should be talking about room events.

> Could you clarify what you mean by that? Update to what? If `MessageEventContent` is indeed only for `m.room.message` subtypes, I don't think we have any enum that encapsulates different event contents. So a new one would need to be defined.

Exactly.

> It's unclear if or why this should be in ruma-client-api.

It shouldn't. This is basically #47, but since there is now a need in such a critical endpoint, I think I'll create just the room event content collection manually before that issue is solved generally.

I'm still unclear though on whether this endpoint can be used only for non-state room events (the equivalent content collection for our current `only::RoomEvent` collection) or for any room event including state events (the equivalent content collection for our current `all::RoomEvent` collection). What's your reading on that?

poljar:
> I'm still unclear though on whether this endpoint can be used only for non-state room events (the equivalent content collection for our current `only::RoomEvent` collection) or for any room event including state events (the equivalent content collection for our current `all::RoomEvent` collection). What's your reading on that?

Correct, only non-state events should be accepted by the server here, though clients of course can try to send anything they like.

The [PUT state](https://matrix.org/docs/spec/client_server/latest#put-matrix-client-r0-rooms-roomid-state-eventtype-statekey) endpoint is used for state events.

jplatte:
I've checked again and actually the spec does define "message events" to be non-state room events [here](https://matrix.org/docs/spec/client_server/latest#types-of-room-events). I'll take this into account in the next breaking change release of ruma-events, and add `collections::only::RoomEventContent` for 0.21.1.

poljar:
They do call them "message events" but there are more message events than `m.room.message`, my assumption that I edited the correct enum comes from this, well and the usage in the `create_message_event` request.

Not sure if we should also rename the `MessageEventContent` so it's clearer that it's only for m.room.message subtypes. Perhaps `RoomMessageContent`. The only way to find that out is to look where else it's used and go to the parent docs list [here](https://docs.rs/ruma-events/0.21.0/ruma_events/room/message/index.html).

jplatte:
I'm not a fan of `RoomMessageContent`. I think the best solution I've got is defining the enums that cover multiple event types in `collections`. Maybe to have separate names the enums there could be called `AnyEvent`, `AnyRoomEvent`, `AnyMessageEvent` and `AnyStateEvent`.

poljar:
> I'm not a fan of `RoomMessageContent`. I think the best solution I've got is defining the enums that cover multiple event types in `collections`. Maybe to have separate names the enums there could be called `AnyEvent`, `AnyRoomEvent`, `AnyMessageEvent` and `AnyStateEvent`.

I'm not sure why you're not a fan of `RoomMessageContent` it maps nicely to the content of `m.room.message`, only the `m` is lost that way. That said, I don't mind the alternatives to argue more for this one though.

You're suggesting names that contain `Event` in the naming, I think we should get rid of the `Event` in cases where we don't put fully build events in the collection, e.g. an event will have an event id (yes yes, to-device events are an exception), sender, etc... That way it's easier to differentiate between the collections, or what collects things a client sends and what a client receives.

jplatte:
Well, maybe you're right. I've thought a bit about it and `RoomMessage` would fit in nicely with getting rid of the current traits and having more generic structs like `ToDevice` and `Stripped` for the top-level event structures.

Sorry for the confusion about `AnyRoomEvent`. The content collection we've been talking about would be called `AnyRoomEventContent`. Or do you consider the `Event` in that name unnecessary or redundant too?

Either way, I've opened #82 about potentially renaming some things and other API changes for the next breaking-change release. Let's continue the discussion there.

Issue #82, opened 2020-05-04
API ideas for 0.22


jplatte:
Here's my current thinking on how the API for ruma-events could change with the next release:

## Get rid of the traits `Event`, `RoomEvent` and `StateEvent`

Instead

* Add new generic structs like `ToDevice` and `Stripped`
  * `Event` with optional room, state fields
  * `NonRoomEvent` (possibly under a different name) with only the mandatory event fields
  * `RoomEvent` with optional state fields
  * `MessageEvent` with only the mandatory room fields
  * `StateEvent` with all fields mandatory
  * potentially others
* … which also means all the generated `Event` types will be dropped
  * because of this, there would no longer be a reason to have `Event` and `EventContent` suffixes. `FullyReadEventContent` can just become `FullyRead`. `EventJson<FullyReadEvent>` becomes `EventJson<Event<FullyRead>>`.
* … we could optionally gate which content types the generic structs (including those that already exist) accept using marker traits `EventContent`, `RoomEventContent` and `StateEventContent`

## Rename event enums

* `all::Event` => `AnyEvent`
* `all::RoomEvent` => `AnyRoomEvent`
* `all::StateEvent` => `AnyStateEvent`
* `only::Event` => `AnyNonRoomEvent`
* `only::RoomEvent` => `AnyMessageEvent`

with this new naming I think these should be exposed from the crate root rather than from a separate `collections` module.

## Expose all the types at the crate root

I'm less certain about this, but I could also imagine exposing all the event content types from the crate root. This would involve another renaming of many event content types, those that are scoped under a submodule like `room`. Examples:

* `fully_read::FullyReadEvent` => `Event<FullyRead>`
* `room::MessageEvent` => `Event<RoomMessage>`
* `room::message::feedback::FeedbackEventContent` => `RoomMessageFeedback`

CC @poljar who contributed some ideas w.r.t. renaming, @iinuwa and @aledomu who also expressed some opinions about potential ruma-events API changes in #ruma:matrix.org

poljar:
Continuing from https://github.com/ruma/ruma-events/pull/78#issuecomment-623349693

> Well, maybe you're right. I've thought a bit about it and `RoomMessage` would fit in nicely with getting rid of the current traits and having more generic structs like `ToDevice` and `Stripped` for the top-level event structures.
> 
[...]

PR #84, opened 2020-05-05
Remove Event, RoomEvent, StateEvent traits  and related macros + codegen

DevinR528:
Starts work on #82


jplatte:
Thanks for working on this! :)

Here's the next steps that I think should probably each be in their own commit:

* Remove the `Event` types, leaving only `EventContent` types
* Add one of the proposed generic event types, for example `StateEvent`, and its corresponding collection type `AnyStateEvent` (a separate private module `state` for this with re-exports at the crate root would probably be a good idea). I think the collection `Deserialize` and `Serialize` can actually be derived, but in a not-so-optimal fashion: We will have to re-declare the "type" string for every collection. Also, it would be nice to have `StateEvent<Something>`'s serialized form also include the type field (which I think isn't currently the case for `ToDevice<Something>` and `Stripped<Something>`).

DevinR528:
Just to clarify "`Event` types" are anything inside of the `Event/RoomEvent/StateEvent` enums, removing the generated `Event` types also? Then in the next commit replace them with the proposed generic types, we'll start with `StateEvent` to see how it goes.

jplatte:
Yes, exactly. Though maybe it would be better to comment out the `collections` module for now instead of deleting it altogether.

jplatte:
I've just set up a new branch, `api22`, to hold the new API as long as it's still incomplete. Could you please open another PR against that branch with the same changes you've already done except

1. not as a WIP PR
2. without removing `fn to_camel_case` and the `kind` field on `RumaEvent` in the macro crate
   * we're going to need the former for other codegen later
   * we're going to need the latter for [this](https://github.com/ruma/ruma-events/issues/82#issuecomment-624206044)

Thanks!

PR #85, opened 2020-05-07
remove Event, RoomEvent and StateEvent traits

DevinR528:


jplatte:
Can you add `#![allow(dead_code)]` to `gen.rs` so CI stops complaining about that?

DevinR528:
Can I add it to `lib.rs` I started removing all the `Event` types and leaving some of the imports and stuff as a place holder (I'm assuming a lot of it will be used once we move forward?).

jplatte:
Yeah, sounds good.

DevinR528:
Ok ignoring the names, this is what were going for right?

src/state/events.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// m.room.topic
+    RoomTopic(),
+
+    /// Any state event that is not part of the specification.
+    CustomState(),
+}
+
+#[derive(Clone, Debug, Serialize)]
+/// To-device event.
```
jplatte, line 65: We generally do this the other way around: Doc comment first, attributes second.

src/state/events.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// A unique key which defines the overwriting semantics for this piece of room state.
+    ///
+    /// This is often an empty string, but some events send a `UserId` to show
+    /// which user the event affects.
+    pub state_key: String,
+    /// Optional previous content for this event.
+    pub previous_content: Option<C>,
+}
+
+pub type AnyStateAliasEvent = StateEvent<AliasesEventContent>;
```
jplatte, line 89: I think having type definitions for each valid combination of generic event struct and content type will needlessly inflate the docs (and the code, unless generated). I think we should just use `StateEvent<AliasesEventContent>` directly.

src/lib.rs
```diff
@@ -169,6 +168,10 @@ pub mod receipt;
 pub mod room;
 pub mod room_key;
 pub mod room_key_request;
+/// The `StateEvent` collection.
+pub mod state {
+    pub mod events;
```
jplatte, line 173: Why `state::events` over just `state`?

jplatte:
For this PR, I was expecting a having only removals, with the new generic structs being added in separate PRs. However it's not that big of a deal, and for the generic structs, yes, this is what I was going for :)

src/state/events.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+#[derive(Clone, Debug, Serialize)]
+/// To-device event.
+pub struct StateEvent<C> {
+    /// Data specific to the event type.
+    pub content: C,
+    /// The type of event.
+    #[serde(rename = "type")]
+    pub event_type: EventType,
```
jplatte, line 71: This seems like something we shouldn't have to store since there is only ever one valid value for the field for each "incarnation" of this type. (as in the value is known at compiletime)

I have a rough idea for how to solve this, but before I rush ahead with that, do you have any ideas?

DevinR528:
I can revert to the previous commit and fix the reviews if you want?

src/lib.rs
DevinR528: That was in case `state` needed to be more than one file. Are we removing the `raw` idea or will there still be serialize/deserialize halves?

jplatte: I'm not a huge fan of the `raw` stuff, but it has some advantages; also we're already changing plenty of things at once. I think it's not going to be that much code that having multiple files for the modules makes sense though. Let's have an inline `raw` module like in `stripped` and `to_device`.

src/state/events.rs
DevinR528: Could we use the `serde(tag = "type")` and somehow flatten the enum into the resulting json?

jplatte:
If that doesn't complicate things for you, sure.

src/state/events.rs
jplatte: Ah yeah, making `AnyStateEvent` an internally tagged enum. I think that would mostly work, yeah :slightly_smiling_face:. My own version is much more complicated, but means event collections actually just become `GenericEvent<AnyGenericEventContent>`. [Does it make any sense to you?](https://gist.github.com/jplatte/18a0997dce8c9357d2bcafd8a78ef546)

jplatte:
To make the diff more manageable, could you, instead of commenting out specific event's tests which don't work, comment out the `mod` declaratiosn for those events?

DevinR528:
Ahh got it sorry I'll have a correct commit in a second.

ruma-events-macros/src/gen.rs
```diff
@@ -213,47 +152,28 @@ impl ToTokens for RumaEvent {
...
-            #[serde(rename = #event_type_name, tag = "type")]
-            pub struct #name {
-                #(#event_fields),*
-            }
+            // #(#attrs)*
+            // #[derive(Clone, Debug, serde::Serialize, ruma_events_macros::FromRaw)]
+            // #[serde(rename = #event_type_name, tag = "type")]
+            // pub struct #name {
+            //     #(#event_fields),*
+            // }
```
jplatte, line 162: I don't think we need to keep this around.

ruma-events-macros/src/gen.rs
```diff
@@ -213,47 +152,28 @@ impl ToTokens for RumaEvent {
...
-                #(#attrs)*
-                #[derive(Clone, Debug, serde::Deserialize)]
-                pub struct #name {
-                    #(#event_fields),*
-                }
+                // #(#attrs)*
+                // #[derive(Clone, Debug, serde::Deserialize)]
+                // pub struct #name {
+                //     #(#event_fields),*
+                // }
```
jplatte, line 176: Same here.

tests/ruma_events_macros.rs
```diff
@@ -1,276 +1,276 @@
-use std::{
```
jplatte, line 1: If you could move this file (e.g. to a directory `tests_backup`), that would make the diff much more readable again (and ensure that nothing is accidentally changed in the comment out > comment back in process).

src/push_rules.rs
```diff
@@ -243,7 +243,7 @@ mod tests {
     use serde_json::{from_value as from_json_value, json, to_value as to_json_value};
```
jplatte, line 243: Please comment out the declaration for `push_rules` instead of commenting out such large portions of the module itself.

src/room/name.rs
```diff
@@ -5,40 +5,7 @@ use std::time::SystemTime;
 use ruma_identifiers::{EventId, RoomId, UserId};
```
jplatte, line 5: Please comment out the declaration for `room::name` instead of commenting out such large portions of the module itself.

src/state/events.rs
jplatte: Since this PR is now only removing stuff, and also the discussion might get quite long, [we'll continue it on #ruma-gsoc:matrix.org](https://matrix.to/#/!EMnjkDIctWzrOMnDuy:matrix.org/$hO3UqlXAEqF6ZnFazfjuRNJ74_F_e2uoqc61-uTGfr0?via=matrix.org).

jplatte:
Okay, now just remove all the `// `s from `tests_backup/ruma_events_macros.rs` and make CI happy, then I'll merge this.

src/room/name.rs
```diff
@@ -288,74 +202,74 @@ mod tests {
...
-            from_json_value::<EventJson<NameEvent>>(json_data)
-                .unwrap()
-                .deserialize()
-                .unwrap()
-                .content
-                .name,
-            name
-        );
-    }
+    // #[test]
```
jplatte, line 205: Why is this commented out? (again?)

DevinR528: This is one of the things I want to work on over the summer, my attention to detail and double checking my work. I will start now.

jplatte:
Looking good now :)

PR #86, opened 2020-05-09
StateEvent collection

DevinR528:
This is without a `raw` version, it would be simple to add raw back in just wanted to see what you thought. I added some notes about `from_parts` and `flatten` attr not working for deserialization.


jplatte:
> it would be simple to add raw back in just wanted to see what you thought

Maybe it actually makes sense to keep `FromRaw` out of the picture as long as we you're not working on codegen yet, considering that would pretty much just be useless copy/pasting.

src/lib.rs
```diff
@@ -214,3 +215,54 @@ impl UnsignedData {
         //   && self.redacted_because.is_none()
     }
 }
+
+/// The base trait that all event content types implement.
+///
+/// Implementing this trait allows content types to be serialized as well as deserialized.
```
jplatte, line 221: Thanks for writing docs right away! :)

src/lib.rs
```diff
@@ -214,3 +215,54 @@ impl UnsignedData {
...
+pub trait EventContent: Sized + Serialize {
+    // TODO When deserializing I couldn't figure out how to get the `event_type`, is
+    // do we need it for validation?
```
jplatte, line 224: We need it for the implementation of `EventContent` for the collection types, otherwise we wouldn't know which variant we're about to deserialize from `content`.

src/lib.rs
```diff
@@ -214,3 +215,54 @@ impl UnsignedData {
...
+        state.end()
+    }
+
+    /// Deserialize a generic state, room or plain event type.
+    pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+    where
+        T: EventContent,
+        D: Deserializer<'de>,
+    {
+        let json = String::deserialize(deserializer)?;
```
jplatte, line 264: Deserializing to a `String` works? I was thinking of a `Deserialize` implementation like shown [here](https://serde.rs/deserialize-struct.html) (probably using the `#[serde(field_identifier)]` attribute that is mentioned there, although I wonder whether it actually works given it's not documented anywhere else).

src/state.rs
```diff
@@ -0,0 +1,199 @@
...
+    // RoomTopic(StateEvent<TopicEventContent>),
+
+    // /// Any state event that is not part of the specification.
+    // CustomState(StateEvent<CustomEventContent>),
+}
+
+/// To-device event.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct StateEvent<C: StateEventContent> {
+    // TODO flatten fails when deserializing for some reason
```
jplatte, line 88: Could you show me the error message?

src/state.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Data specific to the event type.
+    #[serde(flatten, with = "crate::event_content_type_serde")]
+    pub content: C,
+    /// The globally unique event identifier for the user who sent the event.
```
jplatte, line 92: Could you put empty lines between all fields with doc comments (and also between all methods)? Just for easier readability.

```suggestion
    pub content: C,

    /// The globally unique event identifier for the user who sent the event.
```

src/state.rs
```diff
@@ -0,0 +1,199 @@
...
+impl EventContent for AliasesEventContent {
+    fn event_type(&self) -> &str {
+        "m.room.aliases"
+    }
+    fn from_parts(content: &RawValue) -> Result<Self, InvalidEvent> {
+        serde_json::from_str::<AliasesEventContent>(content.get()).map_err(|e| InvalidEvent {
+            kind: InvalidEventKind::Deserialization,
+            message: e.to_string(),
+        })
+        // TODO do we still need this if we can deserialize directly to the type
```
jplatte, line 119: Checking the event type after having built the content is pointless, but once this function has an event type parameter again checking that would potentially be useful. I'm not entirely certain though, maybe we should just ignore it.

src/state.rs
```diff
@@ -0,0 +1,199 @@
...
+        tombstone::TombstoneEventContent,
+        topic::TopicEventContent,
+    },
+    EventContent, EventType, RoomEventContent, StateEventContent,
+};
+
+/// A state event.
+#[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum AnyStateEvent {
```
jplatte, line 33: We don't actually need this enum. Instead, we need an event content collection:

```rust
enum AnyStateEventContent {
    RoomAliases(AliasesEventContent),
    ...
}
```

The event collection you are declaring here is then simply `StateEvent<AnyStateEventContent>`.

src/state.rs
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    use ruma_identifiers::{EventId, RoomAliasId, RoomId, UserId};
+    use serde_json::{from_value as from_json_value, json, to_value as to_json_value};
+
+    use super::{AliasesEventContent, StateEvent};
+    use crate::{EventJson, UnsignedData};
+
+    #[test]
+    fn serialization() {
```
jplatte, line 148: The tests for `StateEvent<AliasesEventContent>` should be in the `room::aliases` module. Here we should have tests for `StateEvent<AnyStateEventContent>`.

src/state.rs
DevinR528: This is what I get
```
---- state::tests::deserialization stdout ----
thread 'state::tests::deserialization' panicked at 'called `Result::unwrap()` on an `Err` value: Error("can only flatten structs and maps", line: 0, column: 0)', src/state.rs:192:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jplatte: Sounds like this is happening because you're trying to deserialize the content as a string, which I was already thinking couldn't work.

DevinR528: Yup I realized that shortly after.

src/lib.rs
DevinR528: I'm working on implementing a custom deserializer for `StateEvent<C: StateEventContent>` it doesn't seem to be terribly difficult we'll see...

jplatte: A custom deserializer? Why?

DevinR528: > I was thinking of a Deserialize implementation like shown here

I used the example from the link, calling it a custom deserializer isn't accurate rather a manually impl'ed `Deserialize`.

src/state.rs
jplatte: Why did you mark this as resolved?

DevinR528: > Here we should have tests for StateEvent<AnyStateEventContent>.

Aren't these testing `StateEvent<AnyStateEventContent>?

jplatte: Ah, you updated them. Sorry for the noise!

src/lib.rs
```diff
@@ -214,3 +215,58 @@ impl UnsignedData {
...
+        state.end()
+    }
+
+    /// Deserialize a generic state, room or plain event type.
+    pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+    where
+        T: EventContent,
+        D: serde::de::Deserializer<'de>,
+    {
+        let json = serde_json::value::Value::deserialize(deserializer)?;
```
jplatte, line 267: Why are you now deserializing to `serde_json::Value` instead of declaring visitors, though?

src/lib.rs
```diff
@@ -214,3 +215,58 @@ impl UnsignedData {
...
+        let ev_type = json.get("type").unwrap().as_str().unwrap();
+        let raw = serde_json::value::to_raw_value(&json).map_err(D::Error::custom)?;
+        T::from_parts(ev_type, &raw).map_err(D::Error::custom)
```
jplatte, line 270: I was expecting the content parameter here to be passed the value of the `content` field, not the whole event object. With that, the individual `from_parts` implementations will also become simpler.

src/lib.rs
```diff
@@ -259,14 +271,124 @@ pub mod event_content_type_serde {
...
+    impl<'de, C: EventContent> Deserialize<'de> for Event<C> {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: serde::Deserializer<'de>,
+        {
+            deserializer.deserialize_struct(
+                "StateEvent",
+                FIELDS,
+                EventVisitor(std::marker::PhantomData),
+            )
```
jplatte, line 305: You could just do this inside the deserialize function above and skip defining a separate struct completely. However, I just realized this custom type is really very helpful. With it, we can skip all of this manual deserialization code and instead derive it as usual (just with `#[serde(skip)]` on the `PhantomData`).

src/state.rs
```diff
@@ -125,18 +125,22 @@ impl EventContent for AnyStateEventContent {
 
     fn from_parts(event_type: &str, content: &RawValue) -> Result<Self, InvalidEvent> {
         fn deserialize_variant<T: StateEventContent>(
```
jplatte, line 127: If you make this a closure, you don't have to pass `ev_type` and `input` explicitly :)

src/lib.rs
```diff
@@ -243,10 +243,22 @@ pub trait StateEventContent: EventContent {}
...
+        ser::SerializeStruct,
+        Serializer,
+    };
+
+    use crate::EventContent;
 
     /// Serialize a generic state, room or plain event type.
+    ///
+    /// Used when serializing the content field of `Event` types like
+    /// so `#[serde(with = "crate::event_content_type_serde)]`.
```
jplatte, line 261:
```suggestion
    /// so `#[serde(with = "crate::event_content_type_serde")]`.
```

src/lib.rs
```diff
@@ -259,14 +271,124 @@ pub mod event_content_type_serde {
     }
 
     /// Deserialize a generic state, room or plain event type.
+    ///
+    /// Used when deserializing the content field of `Event` types like
+    /// so `#[serde(with = "crate::event_content_type_serde)]`.
```
jplatte, line 276:
```suggestion
    /// so `#[serde(with = "crate::event_content_type_serde")]`.
```

src/state.rs
DevinR528: Is there a way to still declare `T: StateEventContent`?

src/lib.rs
DevinR528: Am I missing something
```rust
pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
where
    T: EventContent,
    D: serde::de::Deserializer<'de>,
{
    #[derive(Deserialize)]
    #[serde(bound = "C: EventContent")]
    struct Event<'a, C: EventContent> {
        #[serde(borrow)]
        content: &'a serde_json::value::RawValue,
        event_type: String,
        #[serde(skip)]
        _mk: std::marker::PhantomData<C>,
    }

    let Event {
        event_type,
        content,
        ..
    } = Event::<T>::deserialize(deserializer)?;
    T::from_parts(&event_type, &content).map_err(D::Error::custom)
}
```
gives this error
```
Error("invalid type: newtype struct, expected any valid JSON value", line: 0, column: 0)
```

src/state.rs
DevinR528: is there a way to still do this `let content = T::from_parts(ev_type, input)?;`

src/lib.rs
```diff
@@ -214,3 +215,180 @@ impl UnsignedData {
         //   && self.redacted_because.is_none()
     }
 }
+
+/// The base trait that all event content types implement.
+///
+/// Implementing this trait allows content types to be serialized as well as deserialized.
+pub trait EventContent: Sized + Serialize + serde::de::DeserializeOwned {
```
jplatte, line 222: Event content collections *cannot* implement `Deserialize`, otherwise we wouldn't need any of this.

src/lib.rs
jplatte: I have an idea of why this doesn't work. Try rewriting everything to use `serde_json::Value` for now, instead of `Box<RawValue>`. The existing is wrong though, because it uses `C`'s `Deserialize` implementation, but event content collections shouldn't have one (go through what I wrote in the GSoC room again if you don't know why).

src/state.rs
jplatte: Ah, right, it doesn't work because the function is generic. In that case, nevermind.

src/lib.rs
```diff
@@ -273,122 +273,33 @@ pub mod event_content_type_serde {
...
-    /// so `#[serde(with = "crate::event_content_type_serde)]`.
+    /// so `#[serde(with = "crate::event_content_type_serde")]`.
     pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
     where
         T: EventContent,
         D: serde::de::Deserializer<'de>,
     {
+        use serde_json::value::to_raw_value;
+        #[derive(Deserialize)]
+        struct Event<C: EventContent> {
```
jplatte, line 284: Why does this generic parameter exist, actually?

src/lib.rs
```diff
@@ -214,3 +215,91 @@ impl UnsignedData {
...
+
+/// Marker trait for room events.
+pub trait RoomEventContent: EventContent {}
+
+/// Marker trait for state events.
+pub trait StateEventContent: EventContent {}
+
+/// Serialize all event content types when inside one of `StateEvent<C>`,
+/// `RoomEvent<C>` or `Event<C>`.
+pub mod event_content_type_serde {
```
jplatte, line 243: Could you move this into its own file?

src/state.rs
```diff
@@ -0,0 +1,364 @@
...
+        tombstone::TombstoneEventContent,
+        topic::TopicEventContent,
+    },
+    EventContent, EventType, RoomEventContent, StateEventContent,
+};
+
+/// Serialize all event prev_content types when inside one of `StateEvent<C>`.
+///
+/// `prev_content` is always `Option<C>`.
+pub mod prev_content_type_serde {
```
jplatte, line 34: I hadn't considered `prev_content` before. That does indeed require some special handling. We will need this for types other generic structs that aren't `StateEvent` though (specifically `RoomEvent`, which can be either a message or state event; and `Event`, which is any event). As such, could you make this a submodule of the crate root?

src/state.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Serialize event content types when wrapped in `Option`.
+    ///
+    /// Used when serializing the prev_content field of `StateEvent` types like
+    /// so `#[serde(with = "crate::prev_content_type_serde")]`.
+    pub fn serialize<T, S>(content: &Option<T>, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        T: EventContent,
+        S: Serializer,
+    {
+        // this only runs when Option::Some
```
jplatte, line 56: Could you explain why you added `serde(skip_if_serializing = "Option::is_none")` to the struct rather than checking whether `content` is `Some` or `None` here?

src/state.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Deserialize event content types when wrapped in `Option`.
+    ///
+    /// Used when serializing the prev_content field of `StateEvent` types like
+    /// so `#[serde(with = "crate::prev_content_type_serde")]`.
+    pub fn deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+    where
+        T: EventContent,
+        D: serde::de::Deserializer<'de>,
+    {
+        crate::event_content_type_serde::deserialize(deserializer)
```
jplatte, line 72: This deserializes `content`, while it should check whether `prev_content` exists, and deserialize it if that is the case.

src/state.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Optional previous content for this event.
+    #[serde(
+        flatten,
+        skip_serializing_if = "Option::is_none",
+        with = "prev_content_type_serde"
+    )]
+    pub prev_content: Option<C>,
+}
+
+impl<C: StateEventContent> EventContent for Option<C> {
```
jplatte, line 166: Why does this `impl` exist?

src/state.rs
DevinR528: No good reason, it's way more readable and predictable written out in the `prev_content_serde` mod, I changed it.

src/event_content_serde.rs
```diff
@@ -0,0 +1,60 @@
...
+        #[serde(rename = "type")]
+        event_type: String,
+    }
+
+    let Event {
+        content,
+        event_type,
+        ..
+    } = Event::deserialize(deserializer)?;
+    // Serialize fails without `flatten`, Deserialize fails with flatten without indexing into the RawValue
```
jplatte, line 54: I don't understand this comment. The `Event` type is only used for deserialization. I don't think `#[serde(flatten)]` is required. Can you explain?

src/prev_content_serde.rs
```diff
@@ -0,0 +1,67 @@
...
+pub fn deserialize<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+where
+    T: EventContent,
+    D: serde::de::Deserializer<'de>,
+{
+    #[derive(serde::Deserialize)]
+    struct Event {
+        #[serde(flatten)]
+        content: serde_json::Value,
+        #[serde(rename = "type")]
```
jplatte, line 50: Since the type field is needed for deserialization but will be written by `event_content_serde` in serialization:

```suggestion
        #[serde(rename = "type", skip_serializing)]
```

src/prev_content_serde.rs
```diff
@@ -0,0 +1,67 @@
...
+/// so `#[serde(with = "crate::prev_content_type_serde")]`.
+pub fn deserialize<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+where
+    T: EventContent,
+    D: serde::de::Deserializer<'de>,
+{
+    #[derive(serde::Deserialize)]
+    struct Event {
+        #[serde(flatten)]
+        content: serde_json::Value,
```
jplatte, line 49: Once the general flatten issue is resolved, this should be

```suggestion
        prev_content: Option<serde_json::Value>,
```

src/event_content_serde.rs
```diff
@@ -0,0 +1,60 @@
...
+        #[serde(flatten)]
+        content: serde_json::Value,
+        #[serde(rename = "type")]
+        event_type: String,
+    }
+
+    let Event {
+        content,
+        event_type,
+        ..
```
jplatte, line 52: No longer needed

```suggestion
        ..
```

src/state.rs
```diff
@@ -0,0 +1,312 @@
...
+                _ => unreachable!(
+                    "expected `AnyStateEventContent::RoomAliases found {:?}",
+                    state.prev_content
+                ),
+            }
+        }
+    }
+
+    #[test]
+    #[allow(irrefutable_let_patterns)]
```
jplatte, line 276: This can be removed now that `AnyStateEventContent` has more than one variant, right?

src/state.rs
```diff
@@ -0,0 +1,312 @@
...
+            },
+            "room_id": "!roomid:room.com",
+            "sender": "@carl:example.com",
+            "state_key": "",
+            "type": "m.room.aliases"
+        });
+
+        let state = from_json_value::<StateEvent<AnyStateEventContent>>(json_data).unwrap();
+
+        if let AnyStateEventContent::RoomAliases(content) = state.content {
```
jplatte, line 256: You can use `assert_matches!` for the deserialization tests (like [here](https://github.com/ruma/ruma-events/blob/17faaaa56e08d8e26015d6009f581c7e3e970110/src/room/member.rs#L233)).

src/prev_content_serde.rs
```diff
@@ -0,0 +1,67 @@
...
+    T: EventContent,
+    S: Serializer,
+{
+    // this only runs when Option::Some
+    if content.is_some() {
+        let mut state = serializer.serialize_struct(type_name::<T>(), 1)?;
+        state.serialize_field("prev_content", &content)?;
+        state.end()
+    } else {
+        serializer.serialize_none()
```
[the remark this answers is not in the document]
DevinR528: We do serialize when its `Option::None` I removed the `serde(skip_serializing_if = "Option::is_none")` attribute on `prev_content`, should I replace it then this works
```rust
pub fn serialize<T, S>(opt_content: &Option<T>, serializer: S) -> Result<S::Ok, S::Error>
where
    T: EventContent,
    S: Serializer,
{
    let mut state = serializer.serialize_struct(type_name::<T>(), 1)?;
    state.serialize_field("prev_content", &opt_content)?;
    state.end()
}
```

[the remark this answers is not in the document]
jplatte: Almost. The `serialize_field` call should be optional, only being called with the inner content if it exists.

src/event_content_serde.rs
DevinR528: The comment should actually say `Deserialize fails without 'flatten'...`

Without `flatten` for `Event.content` when deserializing I get an error `Error("missing field 'type'", line: 0, column: 0)`. 

With flatten the `content` field of `Event` is
```
Object(
    {
        "content": Object(
            {
                "aliases": Array(
                    [
                        String(
                            "#somewhere:localhost",
                        ),
                    ],
                ),
            },
        ),
        "prev_content": Object(
            {
                "aliases": Array(
                    [
                        String(
                            "#inner:localhost",
                        ),
                    ],
                ),
            },
        ),
    },
)
```
So I have to pull out `content` (`content.get("content")...`),

jplatte: I don't understand what the `flatten` attribute on `content` has to do with the `type` field. Is that an error message you got before or did you try it out again and got that?

DevinR528: No, I just tried it with everything but this issue resolved. It doesn't make any sense to me either... could it be skipping directly to `AliasesEventContent` instead of `AnyRoomEvent` -> `AliasesEventContent`. I'll push the changes I have so far so you can take a look at them.

jplatte: Okay, I can reproduce. The only thing that comes to my mind that might be causing this is serde's codegen changing between the different versions. I think with two `flatten` members that both want to access the same third field in deserialization, we've hit kind of a special case, and the additional nested `flatten` just fixes things by chance more or less.

jplatte:
Given the issues with `RawValue` and `serde(flatten)`, I think using serde's own codegen for the `Deserialize` implementation for `StateEvent` isn't going to work out. We should probably go back to implementing `Deserialize` (and `Serialize`) [manually](https://serde.rs/deserialize-struct.html) - though let's try to apply this thing serde's docs mention to not make the repetitive code *as* big:

```rust
#[derive(Deserialize)]
#[serde(field_identifier, rename_all = "snake_case")]
enum Field { EventType, Content, OriginServerTs, RoomId, PrevContent, ... }
```

Also, I didn't comment on this before, but keep in mind the [serde attributes](https://github.com/ruma/ruma-events/blob/206a824f3e959ce69f1559d33a63509ef3f39bd9/ruma-events-macros/src/gen.rs#L287) for the event fields (which now will be expressed as regular code instead).

DevinR528:
So "manually" implementing `Deserialize` and `Serialize` for each collection type NOT implementing it for `*EventContent` to deal with the `content` and `prev_content` fields.

DevinR528:
I left the `event_content_serde.rs/prev_content_serde.rs` so the diff would be less I'll remove them once I know this is the direction you want. If this is the right direction the code-gen would be fairly straight forward... seems like at least.

src/state.rs
```diff
@@ -155,6 +160,161 @@ impl RoomEventContent for AnyStateEventContent {}
...
+    where
+        S: Serializer,
+    {
+        let event_type = self.content.event_type();
+
+        let mut state = serializer.serialize_struct("StateEvent", 7)?;
+        state.serialize_field("content", &self.content)?;
+        state.serialize_field("event_id", &self.event_id)?;
+        state.serialize_field("sender", &self.sender)?;
+        state.serialize_field("origin_server_ts", &self.origin_server_ts)?;
```
jplatte, line 174: This is one of the fields I meant re. serde attributes: it should be serialized as a `UInt` rather than a `Duration`.

src/state.rs
```diff
@@ -155,6 +160,161 @@ impl RoomEventContent for AnyStateEventContent {}
...
+        A: MapAccess<'de>,
+    {
+        let mut content: Option<serde_json::Value> = None;
+        let mut event_type: Option<String> = None;
+        let mut event_id: Option<EventId> = None;
+        let mut sender: Option<UserId> = None;
+        let mut origin_server_ts: Option<SystemTime> = None;
+        let mut room_id: Option<RoomId> = None;
+        let mut state_key: Option<String> = None;
+        let mut prev_content: Option<serde_json::Value> = None;
```
jplatte, line 229: It should now once again work to use `Box<RawValue>` over `Value`.

src/state.rs
```diff
@@ -0,0 +1,518 @@
...
+                    if sender.is_some() {
+                        return Err(de::Error::duplicate_field("sender"));
+                    }
+                    sender = Some(map.next_value()?);
+                }
+                Field::OriginServerTs => {
+                    if origin_server_ts.is_some() {
+                        return Err(de::Error::duplicate_field("origin_server_ts"));
+                    }
+                    origin_server_ts = Some(map.next_value()?);
```
jplatte, line 268: We need the reverse conversion from UInt > SystemTime here.

src/state.rs
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    #[test]
+    fn deserialize_aliases_with_prev_content() {
+        let json_data = json!({
+            "content": {
+                "aliases": [ "#somewhere:localhost" ]
+            },
+            "event_id": "$h29iv0s8:example.com",
+            "origin_server_ts": UNIX_EPOCH + Duration::from_millis(1),
```
jplatte, line 418:
```suggestion
            "origin_server_ts": 1,
```

src/state.rs
```diff
@@ -0,0 +1,518 @@
...
+                      "size": 82595,
+                      "w": 800
+                    },
+                    "thumbnail_url": "mxc://matrix.org",
+                    "w": 1011
+                  },
+                "url": "http://www.matrix.org"
+            },
+            "event_id": "$h29iv0s8:example.com",
+            "origin_server_ts": UNIX_EPOCH + Duration::from_millis(1),
```
jplatte, line 468:
```suggestion
            "origin_server_ts": 1,
```

src/state.rs
DevinR528: Ahh now I get it the changes to the JSON below were needed to get the thing to deserialize correctly.

DevinR528:
Should I work on adding all the variants after resolving the issues?

jplatte:
No, having two variants is sufficient for now.

ruma-events-macros/src/gen.rs
```diff
@@ -89,7 +89,7 @@ impl ToTokens for RumaEvent {
             Content::Struct(fields) => {
                 quote! {
                     #[doc = #content_docstring]
-                    #[derive(Clone, Debug, serde::Serialize)]
+                    #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
```
jplatte, line 92: Could you add a TODO comment about removing this later?

src/event_content_serde.rs
```diff
@@ -0,0 +1,58 @@
+//! Serialize all event content types when inside one of `StateEvent<C>`,
```
jplatte, line 1: With the manual `Deserialize` and `Serialize` implementations, this module is no longer needed, right?

src/prev_content_serde.rs
```diff
@@ -0,0 +1,61 @@
+//! Serialize all event prev_content types when inside one of `StateEvent<C>`.
```
jplatte, line 1: Same as for the other serde module.

src/room/aliases.rs
```diff
@@ -2,6 +2,12 @@
 
 use ruma_events_macros::ruma_event;
 use ruma_identifiers::RoomAliasId;
+use serde_json::value::RawValue;
```
jplatte, line 5:
```suggestion
use serde_json::value::RawValue as RawJsonValue;
```

src/room/avatar.rs
```diff
@@ -1,8 +1,13 @@
 //! Types for the *m.room.avatar* event.
 
 use ruma_events_macros::ruma_event;
+use serde_json::value::RawValue;
```
jplatte, line 4:
```suggestion
use serde_json::value::RawValue as RawJsonValue;
```

src/state.rs
```diff
@@ -0,0 +1,520 @@
...
+
+use js_int::UInt;
+use ruma_identifiers::{EventId, RoomId, UserId};
+use serde::{
+    de::{self, Deserialize, Deserializer, Error as _, MapAccess, SeqAccess, Visitor},
+    ser::{Error, SerializeStruct},
+    Serialize, Serializer,
+};
+use serde_json::{
+    value::{to_raw_value as to_raw_json_value, RawValue},
```
jplatte, line 20:
```suggestion
    value::{to_raw_value as to_raw_json_value, RawValue as RawJsonValue},
```

src/state.rs
```diff
@@ -0,0 +1,520 @@
...
+    Value as JsonValue,
```
jplatte, line 21: This import seems to be unused. Maybe you could locally remove the

```rust
#![allow(dead_code)]
#![allow(unused_imports)]
```

to find dead code / unused imports in the newly added code and add them back afterwards. Although it wouldn't be that bad if some slipped through, I just noticed this one by chance.

PR #87, opened 2020-05-15
Move `raw` mod into FromRaw derive

DevinR528:
Move `raw` mod definition into `FromRaw` derive `EventJson` for `EventContent::from_parts`. I also switched `from_parts` from `&RawJsonValue` to `Box<RawJsonValue>` to avoid re-allocating a String and the RawValue.
```rust
fn from_parts(
    event_type: &str,
    content: &serde_json::value::RawValue,
    content: Box<RawJsonValue>,
) -> Result<Self, InvalidEvent>;
```


src/room/aliases.rs
```diff
@@ -26,17 +26,19 @@ impl EventContent for AliasesEventContent {
         "m.room.aliases"
     }
 
-    fn from_parts(event_type: &str, content: &RawJsonValue) -> Result<Self, InvalidEvent> {
+    fn from_parts(event_type: &str, content: Box<RawJsonValue>) -> Result<Self, InvalidEvent> {
         if event_type != "m.room.aliases" {
             return Err(InvalidEvent {
```
DevinR528, line 31: Should I have used `InvalidInput` as opposed to `InvalidEvent`?

jplatte: I would lean towards `serde_json::Error` actually. But please undo the `Box` change, that's got nothing to do with the rest of this PR. We can discuss separately.

ruma-events-macros/src/from_raw.rs
```diff
@@ -15,6 +15,18 @@ pub fn expand_from_raw(input: DeriveInput) -> syn::Result<TokenStream> {
     };
     let ident = &input.ident;
 
+    let content_docstring = format!("The payload for `{}`.", ident);
```
jplatte, line 18: Interesting that these also had documentation before despite being, but `FromRaw` is more general than this doc comment makes it sound. Let's just remove it.

ruma-events-macros/src/from_raw.rs
```diff
@@ -44,5 +56,12 @@ pub fn expand_from_raw(input: DeriveInput) -> syn::Result<TokenStream> {
                 }
             }
         }
+
+        /// "Raw" versions of the event and its content which implement `serde::Deserialize`.
```
jplatte, line 60: Same here, `FromRaw` is not necessarily just for events (in fact, it's now going to be [almost?] exclusively used for event *content* structs). I don't see a point in having doc comments for private generated code, let's remove.

src/room/aliases.rs
DevinR528: Actually it's not that bad `to_owned` works, the only way I could figure it out before was construction a new `RawValue::from_string` then mapping the error.

src/lib.rs
```diff
@@ -144,7 +145,7 @@ pub mod util;
 extern crate self as ruma_events;
 
 pub mod call;
-pub mod custom;
+// pub mod custom;
```
jplatte, line 148: This module should be trivial to fix - just get rid of the existing `raw` module.

src/custom.rs
```diff
@@ -114,12 +49,20 @@ pub(crate) mod raw {
         /// The unique identifier for the user who sent this event.
         pub sender: UserId,
         /// Additional key-value pairs not signed by the homeserver.
-        #[serde(default)]
+        #[serde(skip_serializing_if = "UnsignedData::is_empty")]
```
jplatte, line 52: Why did you remove `default`?

DevinR528: That was only on that field in the `raw` mod should I add them to the non raw type?

jplatte: Yes! Well, I realized these custom types will probably go away before 0.22 is released anyway, but generally for `FromRaw` types, with the generated raw versions, both serialization and deserialization attributes should be present on the struct. serde will ignore the serialization ones on the raw struct which only implements `Deserialize` and ignore the deserialization ones on the "regular" one which only implements `Serialize`.

ruma-events-macros/src/gen.rs
```diff
@@ -90,7 +89,7 @@ impl ToTokens for RumaEvent {
                 // TODO remove serde::Deserialize when this macro actually generates generic events
```
jplatte, line 89: You should remove this now that the `Deserialize` derive is gone ;)

src/lib.rs
jplatte: Resolved - it doesn't actually make sense to update this because it will look totally different with the 0.22 API.

jplatte:
Thanks!

PR #88, opened 2020-05-21
derive StateEventContent

DevinR528:


ruma-events-macros/src/gen.rs
```diff
@@ -88,7 +88,8 @@ impl ToTokens for RumaEvent {
             Content::Struct(fields) => {
                 quote! {
                     #[doc = #content_docstring]
-                    #[derive(Clone, Debug, serde::Serialize, ::ruma_events_macros::FromRaw)]
+                    // TODO we don't want to derive it here do we?
```
jplatte, line 91: We do want to derive it here, however conditionally: Only if `self.event_type` is `StateEvent`. For now, we'll not derive anything in the other cases. Later, we'll use other derive macros for those.

ruma-events-macros/src/lib.rs
```diff
@@ -145,3 +148,13 @@ pub fn derive_from_raw(input: TokenStream) -> TokenStream {
         .unwrap_or_else(|err| err.to_compile_error())
         .into()
 }
+
+/// Generates an implementation of `ruma_events::StateEventContent`. This trait allows event
+/// content types to be used inside of `StateEvent<C: StateEventContent>`.
```
jplatte, line 153: I think the trait's purpose should be described in the trait's docs, not in the derive macro's docs.

```suggestion
/// Generates an implementation of `ruma_events::StateEventContent`.
```

jplatte: Also, maybe mention `RoomEventContent` as well?

src/lib.rs
```diff
@@ -144,26 +144,36 @@ pub mod util;
...
+/// Enums for heterogeneous collections of events.
+// pub mod collections {
+//     pub mod all;
+//     pub mod only;
+
+//     mod raw {
+//         pub mod all;
+//         pub mod only;
+//     }
+// }
```
jplatte, line 158: Seems like the rebase went wrong here? I deleted this module.

src/lib.rs
```diff
@@ -144,26 +144,26 @@ pub mod util;
 // that expect `ruma_events` to exist in the prelude.
 extern crate self as ruma_events;
 
-pub mod call;
+// pub mod call;
```
jplatte, line 147: Why is this PR commenting out a bunch of modules that were not commented out before?

DevinR528: Everything was broken the commit before the latest but now I can uncomment.

ruma-events-macros/Cargo.toml
```diff
@@ -13,7 +13,7 @@ repository = "https://github.com/ruma/ruma-api-macros"
 version = "0.21.2"
 
 [dependencies]
-syn = { version = "1.0.22", features = ["full"] }
+syn = { version = "1.0.22", features = ["full", "extra-traits"] }
```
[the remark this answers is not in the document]
DevinR528: That was so I could `println!` oops

ruma-events-macros/src/state.rs
```diff
@@ -0,0 +1,62 @@
...
+
+enum EventMeta {
+    Type(String),
+}
+
+impl Parse for EventMeta {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        input.parse::<Token![type]>()?;
+        input.parse::<Token![=]>()?;
+        Ok(EventMeta::Type(input.parse::<LitStr>()?.value()))
```
jplatte, line 15: If you take the `.value`, you lose the associated span. Maybe it was me previously suggesting `Type(String)`, but we really should be storing the `LitStr`.

ruma-events-macros/src/state.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+/// Create a `StateEventContent` implementation for a struct
+pub fn expand_state_event(input: DeriveInput) -> syn::Result<TokenStream> {
+    let ident = &input.ident;
+
+    let mut event_type = None;
+    for attr in input.attrs.iter() {
```
jplatte, line 24: I don't think we need to go through all attributes, using the first matching one (`input.attrs.iter().find_map(...)`) should be fine, and would be shorter.

ruma-events-macros/src/state.rs
```diff
@@ -0,0 +1,62 @@
...
+        if attr.path.is_ident("ruma_event") {
+            event_type = Some(attr.parse_args::<EventMeta>()?)
+        }
+    }
+    // TODO: emit compiler error
+    let event_type = match event_type {
+        Some(EventMeta::Type(ev_type)) => ev_type,
+        _ => panic!("no event_type attribute found"),
+    };
```
jplatte, line 33: You can use `.ok_or_else(|| syn::Error::new(...))?` to emit an error here. However, I'm not sure what we should use as the span here. Probably just `Span::call_site()`.

ruma-events-macros/src/state.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let event_content_impl = quote! {
+        impl ::ruma_events::EventContent for #ident {
+            fn event_type(&self) -> &str {
+                #event_type
+            }
+
+            fn from_parts(ev_type: &str, content: Box<serde_json::value::RawValue>) -> Result<Self, InvalidEvent> {
```
jplatte, line 41: For consistency, add leading colons before `serde_json`. Also, there's still an `InvalidEvent` there.

src/room.rs
```diff
@@ -10,22 +10,22 @@ use serde::{Deserialize, Serialize};
 pub mod aliases;
 pub mod avatar;
 // pub mod canonical_alias;
-pub mod create;
+// pub mod create;
```
jplatte, line 13: There's still a bunch of commented-out modules here, do they need to be?

DevinR528: Yeah they have tests that test `*Event`. I could remove the tests?

jplatte: We removed the event structs before this PR. Are you sure the *newly* commented out ones also have tests for those?

ruma-events-macros/src/state.rs
```diff
@@ -1,57 +1,71 @@
...
-    }
-    // TODO: emit compiler error
-    let event_type = match event_type {
-        Some(EventMeta::Type(ev_type)) => ev_type,
-        _ => panic!("no event_type attribute found"),
-    };
+    let event_type = input
+        .attrs
+        .iter()
+        .find_map(|attr| attr.parse_args::<EventMeta>().ok())
```
jplatte, line 37: This is not what I meant. We shouldn't just try to parse any attribute. Maybe this should be `find` + `map`, rather than `find_map`, but the check for `attr.path.is_ident` should remain.

jplatte:
Regarding the lint, let's just have nightly fail (revert the last commit). I'll resolve that issue separately on master and rebase api22 after this PR is merged.

PR #99, opened 2020-05-27
Implement first trybuild tests

DevinR528:
In order to make trybuild happy I had to make InvalidEventKind pub and InvalidEvent's fields pub.


jplatte:
What happens when they're not public? If we need to make them public, we should make them `#[doc(hidden)]` at the same time.

DevinR528:
Without them being usable from another crate the code the macro generates fails (`trybuild` creates another crate in order to run the tests).

src/lib.rs
```diff
@@ -167,7 +167,7 @@ pub mod typing;
 
 pub use self::{
     algorithm::Algorithm,
-    error::{FromStrError, InvalidEvent, InvalidInput},
+    error::{FromStrError, InvalidEvent, InvalidEventKind, InvalidInput},
```
jplatte, line 170: Could you move this out of the re-export statement, under a new one with `#[doc(hidden)]`? I'm not sure whether it will currently be included in documentation (probably not because of `#[doc(hidden)]` on the actual definition), but having it here with the "actually" public re-exports seems a bit confusing.

jplatte:
> Without them being usable from another crate the code the macro generates fails.

Can you be more specific (quote the actual error message)? This seems weird to me, maybe there's another way around it.

DevinR528:
Here is the error message
```bash
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error[E0451]: field `message` of struct `ruma_events::error::InvalidEvent` is private
 --> $DIR/01-sanity-check.rs:4:44
  |
4 | #[derive(Clone, Debug, Serialize, FromRaw, StateEventContent)]
  |                                            ^^^^^^^^^^^^^^^^^ private field
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
```

jplatte:
Oh, I finally get it! It's not trybuild that magically accesses stuff from our error type, it's the `StateEventContent` derive that's being used from outside `ruma_events` that is causing this.

In that case, I don't think just making all those public is the right solution. Instead, could you add a `#[doc(hidden)]` constructor (called `wrong_event_type` or sth. like that) to `InvalidEvent` and use that in the event content derive macros instead of constructing the `InvalidEvent` manually?

ruma-events-macros/src/event_content.rs
```diff
@@ -58,10 +58,11 @@ fn expand_room_event(input: DeriveInput) -> syn::Result<TokenStream> {
                 content: Box<::serde_json::value::RawValue>
             ) -> Result<Self, ::ruma_events::InvalidEvent> {
                 if ev_type != #event_type {
-                    return Err(::ruma_events::InvalidEvent {
-                        kind: ::ruma_events::InvalidEventKind::Deserialization,
-                        message: format!("expected `{}` found {}", #event_type, ev_type),
-                    });
+                    return Err(
+                        ::ruma_events::InvalidEvent::wrong_event_type(
+                            format!("expected `{}` found {}", #event_type, ev_type)
```
jplatte, line 63: Let's move the format call into `wrong_event_type`

src/error.rs
```diff
@@ -12,24 +12,31 @@ use std::{
...
     /// A deserialization error from malformed input.
     Deserialization,
     /// An error occurred validating input according the the matrix spec.
     Validation,
 }
 
 impl InvalidEvent {
+    /// Constructor used in the event content macros.
+    ///
+    /// This is public to allow for `trybuild` tests.
```
jplatte, line 32:
```suggestion
    /// This has to be public to allow the macros to be used outside of ruma-events.
```

PR #101, opened 2020-05-28
event_content_collection!: implement parsing input into enum

DevinR528:


ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,65 @@
...
+/// Create a collection from `RumaCollectionInput.
+pub fn expand_collection(input: RumaCollectionInput) -> syn::Result<TokenStream> {
+    let attrs = &input.attrs;
+    let ident = &input.name;
+    let mut variants = vec![];
+    for (idx, var) in input
+        .events
+        .iter()
+        .map(|lit| {
+            format!(
```
jplatte, line 23: Why are you using `format!` here rather than later interpolating the `to_camel_case` and `to_event_content` output into a `quote!` (or `parse_quote!`) invocation?

ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Create a collection from `RumaCollectionInput.
+pub fn expand_collection(input: RumaCollectionInput) -> syn::Result<TokenStream> {
+    let attrs = &input.attrs;
+    let ident = &input.name;
+    let mut variants = vec![];
+    for (idx, var) in input
+        .events
+        .iter()
+        .map(|lit| {
```
jplatte, line 22: Why do this when constructing the iterator instead of in the loop body?

ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,65 @@
...
+            format!(
+                "{}({})",
+                to_camel_case(lit.value()),
+                to_event_content(lit.value())
+            )
+        })
+        .map(|s| parse_str::<syn::Variant>(&s))
```
jplatte, line 29: Same as previous map: Why not do this in the loop body?

ruma-events-macros/src/collection_parse.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// An array of valid matrix event types. This will generate the variants of the event content type "name".
+    /// There needs to be a corresponding variant in `ruma_events::EventType` for
+    /// this event (converted to a valid Rust-style type name by stripping `m.`, replacing the
+    /// remaining dots by underscores and then converting from snake_case to CamelCase).
+    pub events: Vec<LitStr>,
+}
+
+impl Parse for RumaCollectionInput {
+    fn parse(input: ParseStream<'_>) -> parse::Result<Self> {
+        let attrs = input.call(Attribute::parse_outer)?;
```
jplatte, line 30: Parsing attributes in front of `name:`? Why?

ruma-events-macros/src/collection.rs
jplatte: I think I get it. With this, you turn the strings into idents. I just checked though and we can much improve this to keep spans "automatically" in the conversion if we just change `to_camel_case` and `to_event_content` (see review comment on `to_camel_case`).

ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,65 @@
...
+            #( #variants ),*
+        }
+    };
+
+    Ok(collection)
+}
+
+/// Splits the given `event_type` string on `.` and `_` removing the `m.` then
+/// using only the event name append "EventContent".
+fn to_event_content(name: String) -> String {
```
jplatte, line 55: Same as `to_camel_case`:

```suggestion
fn to_event_content(name: LitStr) -> Ident {
```

ruma-events-macros/src/gen.rs
```diff
@@ -180,7 +180,7 @@ fn populate_state_fields(content_name: Ident, fields: Vec<Field>) -> Vec<Field>
 
 /// Splits the given `event_type` string on `.` and `_` removing the `m.` then
 /// camel casing to give the `EventType` variant.
-fn to_camel_case(name: String) -> String {
+pub(crate) fn to_camel_case(name: String) -> String {
```
jplatte, line 183: This function is only ever called with `lit_str.value()` and we really should be preserving spans for it (as we do in this file [on the master branch](https://github.com/ruma/ruma-events/blob/a3f377ec39490745de667d0916edfd7bc49a4289/ruma-events-macros/src/gen.rs#L83-L84)); why not make that mandatory by taking a `LitStr` and returning an `Ident`:

```suggestion
pub(crate) fn to_camel_case(name: LitStr) -> Ident {
```

Also, since this function is not actually used in this file on the api22 branch, let's move it to `collection.rs`.

ruma-events-macros/src/collection_parse.rs
```diff
@@ -0,0 +1,62 @@
...
+        let mut events = vec![];
+        for item in input.parse::<syn::ExprArray>()?.elems.into_iter() {
+            if let Expr::Lit(ExprLit {
+                lit: Lit::Str(lit_str),
+                ..
+            }) = item
+            {
+                events.push(lit_str);
+            } else {
+                let msg = "values of field `events` are required to be a string literal by `event_content_collection!`";
```
jplatte, line 51: I think we don't need to tell the user of where this error is coming from if there's an error span that points right at it :)

```suggestion
                let msg = "values of field `events` are required to be a string literal";
```

ruma-events-macros/src/collection_parse.rs
```diff
@@ -0,0 +1,62 @@
+//! Details of parsing input for the `event_content_collection` procedural macro.
```
jplatte, line 1: Maybe have this as a sub-module of `collection`, with the definition of `RumaCollectionInput` being part of `collection`?

ruma-events-macros/src/collection_parse.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use syn::{
+    parse::{self, Parse, ParseStream},
+    Attribute, Expr, ExprLit, Ident, Lit, LitStr, Token,
+};
+
+mod kw {
```
jplatte, line 8: To appease clippy:

```suggestion
/// Custom keywords for the `event_content_collection!` macro
mod kw {
```

ruma-events-macros/src/collection_parse.rs
DevinR528: Do we not need to parse them at all? I was imagining for docs.

jplatte: Ah, I see. Well, let's keep it as-is for now then. This will probably disappear once there is one macro to generate all collections at once.

DevinR528: Oh right then it'll just be generated.

ruma-events-macros/src/collection.rs
```diff
@@ -1,40 +1,30 @@
...
-        .map(|s| parse_str::<syn::Variant>(&s))
-        .enumerate()
-    {
-        let mut var = var?;
-
-        let content_docstring = &input.events[idx];
-        var.attrs.push(parse_quote! { #[doc = #content_docstring] });
-        
-        variants.push(var)
+    for lit in input.events.iter() {
```
jplatte, line 17: And now the `let mut variants = vec![];` + `for` loop that fills it can just become

```rust
let variants = input.events.iter().map(|event| {
    // ...
}).collect::<syn::Result<_>>()?;
```

:)

ruma-events-macros/src/collection.rs
```diff
@@ -1,40 +1,30 @@
...
+        let var = to_camel_case(lit);
+        let content = to_event_content(lit);
+        let mut variant: syn::Variant = parse_quote! { #var(#content) };
+
+        let content_docstring = lit;
+        variant
+            .attrs
+            .push(parse_quote! { #[doc = #content_docstring] });
```
jplatte, line 25: Why do this in two separate steps / `parse_quote!` invocations?

ruma-events-macros/src/collection.rs
```diff
@@ -61,5 +53,71 @@ fn to_event_content(name: String) -> String {
     event
         .last()
         .map(|event| format!("{}EventContent", event))
-        .unwrap()
+        .map(|id| Ident::new(&id, span))
+        .expect("must event type must contain more than 'm.'")
```
jplatte: I think this becomes more readable if the `expect` happens right after `.last()`.

ruma-events-macros/src/collection.rs
```diff
@@ -61,5 +53,71 @@ fn to_event_content(name: String) -> String {
...
+            // the name of our collection enum
+            let name: Ident = input.parse()?;
+            input.parse::<Token![,]>()?;
+
+            // events field
+            input.parse::<kw::events>()?;
+            input.parse::<Token![:]>()?;
+            // an array of event names `["m.room.whatever"]`
+            let mut events = vec![];
+            for item in input.parse::<syn::ExprArray>()?.elems.into_iter() {
```
jplatte, line 104: This can also be done by collecting into a `syn::Result`. I think it is more idiomatic to do that in cases like this, and it's even got the performance benefit of pre-allocating (though that's not going to matter here).

ruma-events-macros/src/gen.rs
```diff
@@ -180,12 +180,15 @@ fn populate_state_fields(content_name: Ident, fields: Vec<Field>) -> Vec<Field>
 
 /// Splits the given `event_type` string on `.` and `_` removing the `m.` then
 /// camel casing to give the `EventType` variant.
-pub(crate) fn to_camel_case(name: String) -> String {
+pub(crate) fn to_camel_case(name: &LitStr) -> Ident {
```
jplatte, line 183: You didn't respond to or implement my suggestion to move this to `collection.rs` since it's not being used here since the removal of the event traits. I'm guessing you just missed it?

jplatte:
I don't understand what your latest commit does. In particular, why are you no longer splitting on `_`?

DevinR528:
Ahh right, we have to do it two steps I was thinking of the case `m.what_ever` would have given `Ever` the original way, but you are right I didn't cover the `_` case.

ruma-events-macros/src/gen.rs
DevinR528: Sorry, yeah I didn't see it under the code.

jplatte:
What do you mean with the two steps?

DevinR528:
```rust
fn to_event_content(name: &LitStr) -> Ident {
    let span = name.span();
    let name = name.value();

    assert_eq!(&name[..2], "m.");

    let event = name[2..]
        .split('.')
        .last()
        .unwrap();

    let event = event
        .split('_')
        .collect::<String>();

    let content_str = format!("{}EventContent", event);
    Ident::new(&content_str, span)
}
```

jplatte:
Oh, now I get it. I could imagine re-exporting everything under the crate root with the same names as the variants we have here (so `ruma_events::room::message::MessageEventContent` becomes just `ruma_events::RoomMessageEventContent`), but that's a future API change that we shouldn't be concerned about now. The updated code looks good.

ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,143 @@
...
+            let var = to_camel_case(lit);
+            let content = to_event_content(lit);
+
+            let variant = quote! {
+                #[doc = #content_docstring]
+                #var(#content)
+            };
+            syn::parse(variant.into())
+        })
+        .collect::<syn::Result<Vec<Variant>>>()?;
```
jplatte, line 28: If you put a type annotation of `Vec<_>` on the `let variants` line, you can reduce this to `collect::<syn::Result<_>>()?`, which might be more readable due to less `>`s in a row, but I don't much care either way. Feel free to ignore this.

ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,143 @@
...
+                        ..
+                    }) = item
+                    {
+                        Ok(lit_str)
+                    } else {
+                        let msg = "values of field `events` are required to be a string literal";
+                        Err(syn::Error::new_spanned(item, msg))
+                    }
+                })
+                .collect::<syn::Result<Vec<_>>>()?;
```
jplatte, line 134: Not necessary, just noticed that you can reduce this a small bit further.

```suggestion
                .collect::<syn::Result<_>>()?;
```

ruma-events-macros/src/collection.rs
```diff
@@ -0,0 +1,143 @@
...
+    assert_eq!(&name[..2], "m.");
+    let s = name[2..]
+        .split(&['.', '_'] as &[char])
+        .map(|s| s.chars().next().unwrap().to_uppercase().to_string() + &s[1..])
+        .collect::<String>();
+    Ident::new(&s, span)
+}
+
+/// Details of parsing input for the `event_content_collection` procedural macro.
+pub mod parse {
```
jplatte, line 76: I'm curious (don't just change it back!): why did you now inline this module?

DevinR528: Mostly size and how related they are. I did it somewhat thinking that once `gen.rs` and `parse.rs` are gone it could be moved out again.

jplatte: Okay. I'm wondering why it would be a separate module at all then. But I'm fine merging it as-is.

PR #102, opened 2020-05-29
Replace impl_enum! with strum derive macros

### LINKED ISSUE #90: Replace `impl_enum!` with strum derives
It should be possible to replace it with [EnumString](https://github.com/Peternator7/strum/wiki/Derive-EnumString) and [Display](https://github.com/Peternator7/strum/wiki/Derive-Display) derives.

jbaum98:
Closes #90.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ ruma-identifiers = "0.16.1"
 ruma-serde = "0.2.1"
 serde = { version = "1.0.110", features = ["derive"] }
 serde_json = { version = "1.0.53", features = ["raw_value"] }
+strum = { version = "0.18.0", features = ["derive"] }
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -20,6 +20,8 @@ ruma-identifiers = "0.16.1"
 ruma-serde = "0.2.1"
 serde = { version = "1.0.110", features = ["derive"] }
 serde_json = { version = "1.0.53", features = ["raw_value"] }
+strum = "0.18.0"
```
jplatte, line 23: `strum` has a derive feature like serde (though this is not that well documented unfortunately), let's use that.

jbaum98: Ah, i had a feeling that there was something like that but couldn't find it in the docs.

PR #106, opened 2020-06-01
impl traits needed to deserialize from event_content_collection macro

DevinR528:
Move the EventContent and content marker traits into macro.
Move the raw AnyStateEventContent enum and RawEventContent trait into macro, generate all of the AnyStateEventContent type and impls needed for raw.


jplatte:
Re. the private documentation errors, you can add `#![allow(clippy::missing_docs_in_private_items)]` to the macro crate's crate attributes. I don't think we need to document every private function.

ruma-events-macros/src/collection.rs
```diff
@@ -10,52 +10,173 @@ use parse::RumaCollectionInput;
...
+                #[doc = #event_type_str]
+                #variants(#raw_content)
+            ),*
+        }
+    };
+
+    let raw_event_content_impl = quote! {
+        impl ::ruma_events::RawEventContent for #ident {
+            fn from_parts(event_type: &str, content: Box<::serde_json::value::RawValue>) -> Result<Self, String> {
+                fn deserialize_variant<T: ::ruma_events::RawEventContent>(
```
jplatte, line 106: This `quote!` invocation would be shorter and possibly more readable if it didn't have this local function and instead had the same code inside the repetition. I'm not entirely certain though. What do you think?

ruma-events-macros/src/collection.rs
```diff
@@ -10,52 +10,173 @@ use parse::RumaCollectionInput;
...
+            fn event_type(&self) -> &str {
+                match self {
+                    #( Self::#variants(content) => content.event_type()),*
+                }
+            }
+        }
+    };
+
+    let try_from_raw_impl = quote! {
+        impl ::ruma_events::FromRaw for #ident {
```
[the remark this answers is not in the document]
DevinR528: Is the error type `std::convert::Infallible`?

ruma-events-macros/src/collection.rs
DevinR528: Yea the type signature makes it really complicated looking.

ruma-events-macros/src/collection.rs
```diff
@@ -43,14 +42,28 @@ pub fn expand_collection(input: RumaCollectionInput) -> syn::Result<TokenStream>
...
                 match raw {
-                    #( #variants(c) => Self::#variants(::ruma_events::FromRaw::from_raw(c))),*
+                    #( #variants(c) => Ok(Self::#variants(::ruma_events::TryFromRaw::try_from_raw(c)?))),*
+                }
+            }
+        }
+    };
+
+    let serialize_impl = quote! {
+        impl ::serde::Serialize for #ident {
```
jplatte, line 60: I wonder whether we need this: Shouldn't `#[serde(untagged)]` make serde's `Serialize` derive generate the same code?

DevinR528: Yup it did, I'll go back to that :)

ruma-events-macros/src/collection.rs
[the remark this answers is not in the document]
jplatte: Oh, huh. Well no, that's not going to work for all event types. I'll have to go check what `master`s enums do.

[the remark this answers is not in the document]
jplatte: Ah, they use `String`. Probably good to just replicate that: https://github.com/ruma/ruma-events/blob/master/src/collections/only.rs#L168

ruma-events-macros/src/collection.rs
```diff
@@ -45,13 +45,21 @@ pub fn expand_collection(input: RumaCollectionInput) -> syn::Result<TokenStream>
...
+            type Err = String;
 
             fn try_from_raw(raw: Self::Raw) -> Result<Self, Self::Err> {
                 use raw::#ident::*;
 
                 match raw {
-                    #( #variants(c) => Ok(Self::#variants(::ruma_events::TryFromRaw::try_from_raw(c)?))),*
+                    #( #variants(c) => {
+                            let content = ::ruma_events::TryFromRaw::try_from_raw(c)
+                                .map_err(|e: <#content as ::ruma_events::TryFromRaw>::Err| e.to_string());
```
jplatte, line 56:
```suggestion
                                .map_err(|e: <#content as ::ruma_events::TryFromRaw>::Err| e.to_string())?;
```

ruma-events-macros/src/collection.rs
```diff
@@ -45,13 +45,21 @@ pub fn expand_collection(input: RumaCollectionInput) -> syn::Result<TokenStream>
...
+                                // without this ^^^^^^^^^^^ the compiler fails to infer the type
+                            Ok(Self::#variants(
+                                content?
+                            ))
```
jplatte, line 60:
```suggestion
                            Ok(Self::#variants(content))
```

PR #107, opened 2020-06-02
implement MessageEvent and AnyMessageEventContent

DevinR528:


ruma-events-macros/src/collection.rs
```diff
@@ -72,9 +72,12 @@ pub fn expand_collection(input: RumaCollectionInput) -> syn::Result<TokenStream>
 
         #event_content_impl
 
-        impl RoomEventContent for AnyStateEventContent {}
+        // TODO should these not be generated? or some kind of conditional code-gen
```
jplatte, line 75: These should be conditional. `AnyStateEventContent` shouldn't implement `MessageEventContent` and `AnyMessageEventContent` shouldn't implement `StateEventContent`.

I think you could do something like

```rust
static MARKER_TRAITS: Lazy<BTreeMap<Ident, Vec<Path>>> = Lazy::new(|| btreemap! {
    parse_quote!(AnyStateEventContent) => vec![parse_quote!(::ruma_events::MessageEventContent), ...],
    ...
});
```

using `Lazy` from the `once_cell` crate and `btreemap!` from the `maplit` crate.

Or you could have the list of marker traits to implement as part of the macro invocation, but I think the final macro will probably not contain anything like that and instead have some info about the kinds of events hardcoded in the macro crate (but more than just the marker traits).

src/message.rs
```diff
@@ -0,0 +1,428 @@
...
+};
+
+use crate::{MessageEventContent, RawEventContent, RoomEventContent, TryFromRaw, UnsignedData};
+use ruma_events_macros::event_content_collection;
+
+event_content_collection! {
+    /// A message event.
+    name: AnyMessageEventContent,
+    // TODO are these ok to use here or are there specific events I should be using?
+    events: ["m.room.aliases", "m.room.avatar"]
```
jplatte, line 23: These are state events. Message events are any that derive `MessageEventContent` (from those that are still declared using `ruma_event!`, that's those that are declared as `kind: RoomEvent`).

jplatte: The most important one, `m.room.message`, is currently commented out. Let's use the four call events for now.

DevinR528:
Oops I started to implement the call event content stuff then realized you already had. Should be good now.

jplatte:
You could have used those even without them being converted to no longer use `ruma_event!` AFAICT.

DevinR528:
I couldn't get the `Lazy<BTreeMap...` to work as none of the syn types or token stream are `Sync` so I went with a `match` hopefully that works (definitely not as nice as the btreemap).

ruma-events-macros/src/collection.rs
```diff
@@ -139,17 +153,18 @@ fn to_event_content_path(
...
         .split('_')
         .map(|s| s.chars().next().unwrap().to_uppercase().to_string() + &s[1..])
         .collect::<String>();
 
-    let module = Ident::new(event_str, span);
     let content_str = Ident::new(&format!("{}EventContent", event), span);
+    let path = path.iter().map(|s| Ident::new(s, span));
     syn::parse_quote! {
-        ::ruma_events::room::#module::#content_str
+        ::ruma_events::#( #path::)*#content_str
```
jplatte, line 167: I *think* this should work too (and is easier to read IMO):

```suggestion
        ::ruma_events::#(#path)::*::#content_str
```

ruma-events-macros/src/collection.rs
```diff
@@ -161,17 +176,18 @@ fn to_raw_event_content_path(
...
         .split('_')
         .map(|s| s.chars().next().unwrap().to_uppercase().to_string() + &s[1..])
         .collect::<String>();
 
-    let module = Ident::new(event_str, span);
     let content_str = Ident::new(&format!("{}EventContent", event), span);
+    let path = path.iter().map(|s| Ident::new(s, span));
     syn::parse_quote! {
-        ::ruma_events::room::#module::raw::#content_str
+        ::ruma_events::#( #path::)*raw::#content_str
```
jplatte, line 190: Same here:

```suggestion
        ::ruma_events::#(#path)::*::raw::#content_str
```

src/call/answer.rs
```diff
@@ -9,7 +9,7 @@ use super::SessionDescription;
 /// This event is sent by the callee when they wish to answer the call.
 #[derive(Clone, Debug, Serialize, FromRaw, MessageEventContent)]
 #[ruma_event(type = "m.call.answer")]
-pub struct AnswerEventContenet {
+pub struct AnswerEventContent {
```
jplatte, line 12: Thanks for fixing this :D

src/message.rs
```diff
@@ -0,0 +1,482 @@
...
+    Serialize, Serializer,
+};
+
+use crate::{MessageEventContent, RawEventContent, RoomEventContent, TryFromRaw, UnsignedData};
+use ruma_events_macros::event_content_collection;
+
+event_content_collection! {
+    /// A message event.
+    name: AnyMessageEventContent,
+    events: ["m.call.answer", "m.call.invite", "m.call.hangup", "m.call.candidates", "m.sticker"]
```
jplatte, line 22: I think it's time to put each of these onto its own line (and add a trailing comma to `m.sticker` then, let's see whether that breaks the parsing code!).

zzz.rs
```diff
@@ -0,0 +1,603 @@
+mod message {
```
jplatte, line 1: I'm guessing you didn't mean to commit this? ;)

DevinR528: I knew I'd forget one of these times shoot! ;)

ruma-events-macros/src/collection.rs
```diff
@@ -164,7 +164,7 @@ fn to_event_content_path(
     let content_str = Ident::new(&format!("{}EventContent", event), span);
     let path = path.iter().map(|s| Ident::new(s, span));
     syn::parse_quote! {
-        ::ruma_events::#( #path::)*#content_str
+        ::ruma_events::#( #path)::*::#content_str
```
jplatte, line 167: Please write either `#( #path )::*` or `#(#path)::*` (spaces inside braces on both sides, or none).

PR #108, opened 2020-06-03
State event

DevinR528:
I actually did have to squash a few commits :).


ruma-events-macros/src/events.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+    };
+
+    let enum_variants = fields.iter().map(|field| {
+        let name = field.ident.as_ref().unwrap();
+        println!("{:?}", name);
+        to_camel_case(name)
+    }).collect::<Vec<_>>();
+
+    let serialize_impl = quote! {
```
jplatte, line 99: Why is this called `serialize_impl`? It's the main logic for *de*serialization.

ruma-events-macros/src/events.rs
```diff
@@ -0,0 +1,263 @@
...
+    let deserialialize_impl = quote! {
+        impl<'de, C> ::serde::de::Deserialize<'de> for StateEvent<C>
+        where
+            C: ::ruma_events::RawEventContent,
+        {
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: ::serde::de::Deserializer<'de>,
+            {
+                deserializer.deserialize_map(StateEventVisitor(::std::marker::PhantomData))
```
jplatte, line 88: I think it would be nice to interpolate what's currently named `serialize_impl` before this line. It's never used outside of this function and if it's a separate `quote!` invocation, there's also no concerns about indentation depth.

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{
+    Data, DataStruct, DeriveInput, Field, Fields, FieldsNamed, GenericParam, Ident, TypeParam,
+};
+
+/// Derive `Event` macro code generation.
+///
+// TODO better doc comment
+pub fn expand_state_event(input: DeriveInput) -> syn::Result<TokenStream> {
```
jplatte, line 12:
```suggestion
pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
```

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+            if name == "prev_content" {
+                quote! {
+                    if let Some(content) = self.prev_content.as_ref() {
+                        state.serialize_field("prev_content", content)?;
+                    }
+                }
+            } else if name == "origin_server_ts" {
+                quote! {
+                    let time_since_epoch = self.origin_server_ts.duration_since(::std::time::UNIX_EPOCH).unwrap();
+                    let timestamp = match ::js_int::UInt::try_from(time_since_epoch.as_millis()) {
```
jplatte, line 75: You can use `.map_err` + `?` here instead of a `match`, can't you?

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+                    };
+                    state.serialize_field("origin_server_ts", &timestamp)?;
+                }
+            } else if name == "unsigned" {
+                quote! {
+                    if !self.unsigned.is_empty() {
+                        state.serialize_field("unsigned", &self.unsigned)?;
+                    }
+                }
+            }else {
```
jplatte, line 87: Please fix the formatting here.

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+                #( #serialize_fields )*
+                state.end()
+            }
+        }
+    };
+
+    let raw_mod = expand_raw_state_event(&input, fields)?;
+
+    Ok(quote! {
+
```
jplatte, line 118:
```suggestion
```

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+            } else if name == "origin_server_ts" {
+                quote! {
+                    let origin_server_ts = origin_server_ts
+                        .map(|time| ::std::time::UNIX_EPOCH + ::std::time::Duration::from_millis(time.into()))
+                        .ok_or_else(|| ::serde::de::Error::missing_field("origin_server_ts"))?;
+                }
+            } else if name == "unsigned" {
+                quote! { let unsigned = unsigned.unwrap_or_default(); }
+            }else {
```
jplatte, line 193: Format please.

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+    // TODO would all these `iter().maps(` be better as a `for loop` pushing to multiple vecs?
+    let enum_variants = fields
+        .iter()
+        .map(|field| {
+            let name = field.ident.as_ref().unwrap();
+            to_camel_case(name)
+        })
+        .collect::<Vec<_>>();
+
+    let deserialize_variables = fields
```
jplatte, line 151: Since we're only varying the type inside `Option` here, how about moving the `let mut #name: Option<` + `> = None;` parts into the interpolation further down?

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut gen = input.generics.clone();
+        for p in &mut gen.params {
+            if let GenericParam::Type(TypeParam { bounds, .. }) = p {
+                bounds.clear();
+            }
+        }
+        gen
+    };
+
+    // TODO would all these `iter().maps(` be better as a `for loop` pushing to multiple vecs?
```
jplatte, line 142: If `itertools` had `iunzip!` I'd recommend that, but unfortunately it's not a thing ^^

Do whatever you feel is more readable :)

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+
+                        while let Some(key) = map.next_key()? {
+                            match key {
+                                Field::Type => {
+                                    if event_type.is_some() {
+                                        return Err(::serde::de::Error::duplicate_field("type"));
+                                    }
+                                    event_type = Some(map.next_value()?);
+                                }
+                                #(
```
jplatte, line 251: The `content` field also has to be present for any event, so you can treat it like the `type` field here and remove the branches for it above (unless that actually makes things more complicated).

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+
+                        let event_type = event_type.ok_or_else(|| ::serde::de::Error::missing_field("type"))?;
+                        #( #ok_or_else_fields )*
+
+                        Ok(#ident {
+                            #( #field_names ),*
+                        })
+                    }
+                }
+                deserializer.deserialize_map(EventVisitor(::std::marker::PhantomData))
```
jplatte, line 270:
```suggestion
                }

                deserializer.deserialize_map(EventVisitor(::std::marker::PhantomData))
```

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+            }
+        }
+    };
+
+    let raw_docs = format!("The raw version of {}, allows for deserialization.", ident);
+    Ok(quote! {
+
```
jplatte, line 277:
```suggestion
```

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,302 @@
...
+            pub struct #ident #generics {
+                #( #fields ),*
+            }
+
+            #deserialize_impl
+        }
+    })
+}
+
+/// CamelCase's a field ident like "foo_bar" to "FooBar".
```
jplatte, line 292: This comment seems inconsistent with splitting on both `.` and `_` below.

tests/event.rs
```diff
@@ -0,0 +1,6 @@
+#[test]
+fn ui() {
+    let t = trybuild::TestCases::new();
+    // t.pass("tests/ui/04-event-sanity-check.rs"); // rustc overflows when compiling this?
```
jplatte, line 4: An overflow? Of what kind? Let's discuss this in the ruma-gsoc chat.

ruma-events-macros/src/event.rs
DevinR528: Is this just putting a space after the bracket? sorry the diff makes it look like 2

ruma-events-macros/src/event.rs
DevinR528: Ah right since it's only camel casing field names I should only be splitting on `_`?

ruma-events-macros/src/event.rs
DevinR528: Yeah I've encountered a few situations where `unzip` would have been nice. I guess I'll just leave it, a for loop may be fewer lines of code but more convoluted.

jplatte: Well `unzip` exists, it's in `std` as an `Iterator` method. But it only does two collections.

ruma-events-macros/src/event.rs
jplatte: Yeah the diff is not very useful. Just one, anything else would get squashed by rustfmt anyway (in regular code, that is. probably not inside quote :roll_eyes:).

ruma-events-macros/src/event.rs
jplatte: I haven't actually checked where this is used but if it's just for camel casing then yes.

ruma-events-macros/src/event.rs
DevinR528: In order to do this we would have to have a `.filter` to remove the `content` `prev_content` and `origin_server_ts` then do a `iter().flat_map(` to separate the types that aren't as declared in the struct?
```rust
let deserialize_diff_type_variables = fields
        .iter()
        .flat_map(|field| {
            let name = field.ident.as_ref().unwrap();
            if name == "content" {
                Some(quote! { let mut content: Option<Box<::serde_json::value::RawValue>> = None; }
            } else if name == "prev_content" {
                Some(quote! { let mut prev_content: Option<Box<::serde_json::value::RawValue>> = None; }
            } else if name == "origin_server_ts" {
                Some(quote! { let mut origin_server_ts: Option<::js_int::UInt> = None; }
            } else {
                None
            }
        })
        .collect::<Vec<_>>();

let deserialize_variables = fields
        .iter()
        .flat_map(|field| {
            let name = field.ident.as_ref().unwrap();
            if name == "content" {
                || name == "prev_content" {
                || name == "origin_server_ts" 
           {
                None
            } else {
                Some(name)
            }
        })
        .collect::<Vec<_>>();
```
or am I missing something obvious (highly possible)?

jplatte: I don't really get what you're saying. `deserialize_variables` doesn't need to exist AFAICT. I think you should be able to rename it to `deserialize_var_types`, do roughly the same map as now but only returning the type (minus `Option`) the local variable should have. Then you should be able to change `#( #deserialize_variables )*` below to roughly

```rust
#(
    let mut #field_names: Option<#deserialize_var_types> = None;
)*
```

ruma-events-macros/src/event.rs
DevinR528: You would still have to check for it to filter it out, I added a check first thing when destructuring the `DeriveInput` that will error if no `content` field is present how's that?

jplatte: Sounds like a good approach. I'll have a look at the code.

ruma-events-macros/src/event.rs
```diff
@@ -0,0 +1,303 @@
...
+            let name = field.ident.as_ref().unwrap();
+            if name == "prev_content" {
+                quote! {
+                    if let Some(content) = self.prev_content.as_ref() {
+                        state.serialize_field("prev_content", content)?;
+                    }
+                }
+            } else if name == "origin_server_ts" {
+                quote! {
+                    let time_since_epoch = self.origin_server_ts.duration_since(::std::time::UNIX_EPOCH).unwrap();
```
jplatte, line 81: These lines are getting very long. Can you re-format to stay within 100 chars line width?

ruma-events-macros/src/event.rs
jplatte: The TODO comment is still there.

DevinR528:
I did look into it a bit more the overflow happens on every rust (stable, beta, nightly) and `recursion_limit` didn't help (it is a more helpful compiler message anyways so it wasn't that) should I investigate more or just leave it for now. I kinda wonder if it's `trybuild` doing something weird?

tests/ui/06-no-content-field.stderr
```diff
@@ -0,0 +1,5 @@
+error: struct must contain a `content` field
+  --> $DIR/06-no-content-field.rs:10:5
+   |
+10 |     pub not_content: C,
```
jplatte, line 4: I think it would make more sense for this have the `call_site()` span. Pointing at the first or all fields when the content field is missing seems weird, and with `call_site()` it's clear which macro expects there to be a `content` field.

tests/ui/05-named-fields.stderr
```diff
@@ -0,0 +1,5 @@
+error: `Event` only supports named fields
+ --> $DIR/05-named-fields.rs:6:44
+  |
+6 | pub struct StateEvent<C: StateEventContent>(C)
```
jplatte, line 4: *Maybe* same thing here with the span. I'm not sure. This one seems less bad.

DevinR528: How about this
```rust
return Err(syn::Error::new_spanned(
    fields, // still showing the struct but being more explicit about the derive?
    "the `Event` derive only supports named fields",
));
```

jplatte: Sure.

PR #109, opened 2020-06-04
Box optional sub-structs in avatar and message events

jplatte:
… to heavily decrease the size of some types.
Most importantly: MessageEventContent goes from 640 to 104 bytes in size.

Kinda scary how I didn't have to update any tests. We should definitely add tests where these sub-structs are `Some(...)`.


jplatte:
@poljar You got an opinion on this?

Also CC @sparky8251, @DevinR528: are you using any of the fields that are now `Box`es with this PR?

poljar:
Looks good to me, no objections from my side.

DevinR528:
Looks good, I haven't dealt with any of those variants yet.

PR #111, opened 2020-06-05
Remove raw mod and TryFromRaw/FromRaw, add Deserialize to derive for event content

DevinR528:
`EventJson::deserialize` has bound for DeserializeOwned and `EventJson::deserialize_content` has a bound of `T: EventContent` now, that seemed right to me right? The 9 failing tests are the `key::verification::start...` tests and `room::name...` tests do you want me to reimplement the checks with methods/types where possible?


src/json.rs
```diff
@@ -70,26 +60,18 @@ where
...
 impl<T: EventContent> EventJson<T>
 where
-    T::Raw: RawEventContent,
+    T: EventContent,
 {
     /// Try to deserialize the JSON as event content
     pub fn deserialize_content(self, event_type: &str) -> Result<T, InvalidEvent> {
-        let raw_content = match T::Raw::from_parts(event_type, self.json) {
-            Ok(raw) => raw,
+        match T::from_parts(event_type, self.json) {
```
jplatte, line 67: This match can be replaced by `map_err`.

src/json.rs
```diff
@@ -102,7 +84,7 @@ impl<T: Serialize> From<&T> for EventJson<T> {
 
 // Without the `TryFromRaw` bound, this would conflict with the next impl below
 // We could remove the `TryFromRaw` bound once specialization is stabilized.
-impl<T: Serialize + TryFromRaw> From<T> for EventJson<T> {
+impl<T: Serialize + EventContent> From<T> for EventJson<T> {
```
jplatte, line 87: This bound makes this `impl` no longer usable for the full event structs. Instead of replacing `TryFromRaw` here, just remove it altogether. This means removing the conflicting `From<RawJsonValue>` implementation. To not lose that fast path, we then need a dedicated constructor (`from_json`) that does the same thing.

src/room/name.rs
```diff
@@ -13,55 +14,74 @@ use crate::{InvalidInput, TryFromRaw, UnsignedData};
 #[ruma_event(type = "m.room.name")]
 pub struct NameEventContent {
     /// The name of the room. This MUST NOT exceed 255 bytes.
-    pub(crate) name: Option<String>,
+    #[serde(default, deserialize_with = "room_name")]
```
DevinR528, line 17: I can just revert this file if you want, is this an ok solution to validating the type?

src/key/verification/start.rs
```diff
@@ -407,121 +407,127 @@ mod tests {
...
-        });
-        let error = from_json_value::<EventJson<StartEvent>>(json_data)
-            .unwrap()
-            .deserialize()
-            .unwrap_err();
-
-        assert!(error.message().contains("key_agreement_protocols"));
-        assert!(error.is_validation());
-    }
+    // TODO this fails because the error is a Validation error not deserialization?
```
jplatte, line 410: Can you use block comments for a smaller diff please?

jplatte, line 410: Yeah, these tests are for the `TryFromRaw` verification you have removed.

src/room/name.rs
jplatte: You seem to have duplicated the verification here. Either the verification should be in `NameEventContent`s constructor & `Deserialize` implementation (using `#[serde(deserialize_with)]` as you've done here), or it should be done through a separate type, in which case the field can be properly public and `NameEventContent::new` should have a `RoomName` parameter rather than `String`.

src/key/verification/start.rs
DevinR528: Should I just remove the test or will someone add back that error?

jplatte: Let's keep it in. We might want to re-introduce those checks.

PR #112, opened 2020-06-06
Remove src/from_raw.rs and ruma-events-macro/src/from_raw.rs

DevinR528:


ruma-events-macros/src/gen.rs
```diff
@@ -83,7 +83,7 @@ impl ToTokens for RumaEvent {
             Content::Struct(fields) => {
                 quote! {
                     #[doc = #content_docstring]
-                    #[derive(Clone, Debug, ::serde::Serialize, ::ruma_events_macros::FromRaw)]
```
jplatte, line 86: We need to derive `Deserialize` now.

PR #113, opened 2020-06-07
Implement PresenceEvent and EphemeralEvent

DevinR528:
Add derive for `PresenceEventContent` and create struct `AnyPreseenceEventContent` since there is only one content type.
Add derive for `EphemeralEvent` and create enum `AnyEphemeralEventContent`, convert receipt, and typing to use `derive(EphemeralEventContent)` over `ruma_api!`.


src/lib.rs
```diff
@@ -217,6 +218,15 @@ pub trait EventContent: Sized + Serialize {
     fn from_parts(event_type: &str, content: Box<RawJsonValue>) -> Result<Self, String>;
 }
 
+/// Marker trait for the content of a ephemeral event.
+pub trait EphemeralEventContent: EventContent {}
+
+/// Marker trait for the content of a presence event.
+pub trait PresenceEventContent: EventContent {}
```
jplatte, line 225: We don't actually need this.

src/presence.rs
```diff
@@ -1,46 +1,56 @@
...
 }
 
-pub use ruma_common::presence::PresenceState;
+/// Informs the room of members presence.
+///
+/// This is the only event content a `PresenceEvent` can contain as it's
+/// `content` field.
+#[derive(Clone, Debug, Deserialize, Serialize, PresenceEventContent)]
+#[ruma_event(type = "m.presence")]
+pub struct AnyPresenceEventContent {
```
jplatte, line 30:
```suggestion
pub struct PresenceEventContent {
```

ruma-events-macros/src/collection.rs
```diff
@@ -16,6 +16,9 @@ fn marker_traits(ident: &Ident) -> TokenStream {
             impl ::ruma_events::RoomEventContent for #ident {}
             impl ::ruma_events::MessageEventContent for #ident {}
         },
+        "AnyEphemeralEventContent" => quote! {
```
jplatte, line 19:
```suggestion
        "AnyEphemeralRoomEventContent" => quote! {
```

src/ephemeral.rs
```diff
@@ -0,0 +1,173 @@
...
+
+event_content_collection! {
+    /// A ephemeral event.
+    name: AnyEphemeralEventContent,
+    events: [ "m.typing", "m.receipt" ]
+}
+
+/// Ephemeral event.
+#[derive(Clone, Debug, Event)]
+pub struct EphemeralEvent<C: EphemeralEventContent> {
```
jplatte, line 28:
```suggestion
pub struct EphemeralRoomEvent<C: EphemeralRoomEventContent> {
```

src/lib.rs
DevinR528: And make `PresenceEvent` non generic? I figured out a better way to do this in the macro.

jplatte: Yes, `PresenceEvent` should not be generic. I think I also said so before you started working on this over in the Matrix room.

DevinR528: Ah sorry I did see that in the Matrix room I wasn't trying to be an ass :disappointed_relieved:  I had a comment that github lost (I never submitted the review comment) about doing it with generic keeping less "cheating" branches (`if input.ident == "PresenceEvent ...`) out of the macro. So ignore all of that sorry about that.

DevinR528:
Do you want `PresenceEvent` in it's own file?

src/content_enums.rs
```diff
@@ -34,3 +34,9 @@ event_content_enum! {
         "m.room.topic",
     ]
 }
+
+event_content_enum! {
+    /// A ephemeral event.
```
jplatte, line 39: Missing "room", also "an" instead of "a".

src/lib.rs
```diff
@@ -219,6 +219,9 @@ pub trait EventContent: Sized + Serialize {
     fn from_parts(event_type: &str, content: Box<RawJsonValue>) -> Result<Self, String>;
 }
 
+/// Marker trait for the content of a ephemeral event.
```
jplatte, line 222: 'an'

jplatte:
It should be part of the presence module, like it is now.

PR #114, opened 2020-06-08
Remove unused imports and fix Error trait not in scope

DevinR528:


src/event_kinds.rs
```diff
@@ -2,7 +2,6 @@ use std::{convert::TryFrom, time::SystemTime};
 
 use ruma_events_macros::Event;
 use ruma_identifiers::{EventId, RoomId, UserId};
-use serde::ser::Error;
```
jplatte, line 5: A few lines above this, there's also a `TryFrom` import that is not used directly.

ruma-events-macros/src/event.rs
```diff
@@ -164,6 +166,8 @@ fn expand_deserialize_event(
             }
         } else if name == "origin_server_ts" {
             quote! {
+                use ::std::convert::TryFrom as _;
```
jplatte, line 169: This seems disconnected from the actual use (this `quote!` invocation doesn't seem to use `TryFrom`). Could you just update `Type::try_from` to `::std::convert::TryFrom::try_from` if type inference can figure out the types, or `<Type as ::std::convert::TryFrom>::try_from` otherwise?

PR #116, opened 2020-06-08
Add RedactionEvent and convert RedactionEventContent to struct def

DevinR528:
Do you want me to move the `RedactionEventContent` into the `src/redaction.rs`  or should I move `RedactionEvent` into `event_kinds.rs`?


ruma-events-macros/src/event.rs
```diff
@@ -71,6 +72,10 @@ pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
         quote! {
             "m.presence";
         }
+    } else if is_redaction_event {
```
jplatte, line 75: Ah, so actually presence and redaction event content should also implement `EventContent` (and `RedactionEventContent` should additionally implement `RoomEventContent`). Maybe we do need `EventContent` and `RoomEventContent` derives.

Then these special cases will not be required anymore.

ruma-events-macros/src/event.rs
```diff
@@ -105,6 +111,7 @@ pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
 
 fn expand_deserialize_event(
     is_presence_event: bool,
+    is_redaction_event: bool,
```
jplatte, line 114: I think you should be able to collapse these into one flag `is_generic_event` once `PresenceEventContent` and `RedactionEventContent` implement `EventContent`. Or maybe you can simplify the code even further.

ruma-events-macros/src/event.rs
```diff
@@ -67,23 +67,18 @@ pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
...
-
     let serialize_impl = quote! {
         impl #impl_gen ::serde::ser::Serialize for #ident #ty_gen #where_clause {
             fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
             where
                 S: ::serde::ser::Serializer,
             {
                 use ::serde::ser::{SerializeStruct as _, Error as _};
+                // calling event_type <type as EventContent>::event_type requires
+                // all the generic structs to also be bound by Sized
```
jplatte, line 78: I don't understand this comment. This part of the code doesn't deal with bounds. If you have the type, `<Type as ::ruma_events::EventContent>::event_type` should definitely work, but even without that, since it's a method (it has a `self` parameter) and not an associated function, type inference should always work if you use `::ruma_events::EventContent::event_type(self)`.

ruma-events-macros/src/event.rs
```diff
@@ -8,7 +8,7 @@ use syn::{Data, DataStruct, DeriveInput, Field, Fields, FieldsNamed, Ident};
 pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
     let ident = &input.ident;
     let (impl_gen, ty_gen, where_clause) = input.generics.split_for_impl();
-    let is_presence_event = ident == "PresenceEvent";
+    let is_presence_or_redaction = ident == "PresenceEvent" || ident == "RedactionEvent";
```
jplatte, line 11: Can we check whether the input type has any generics instead?

ruma-events-macros/src/event_content.rs
```diff
@@ -23,7 +23,8 @@ impl Parse for EventMeta {
     }
 }
 
-fn expand_event_content(input: DeriveInput) -> syn::Result<TokenStream> {
+/// Create a `EventContent` implementation for a struct.
```
jplatte, line 26: Grammar.

ruma-events-macros/src/event.rs
DevinR528: :facepalm:

ruma-events-macros/src/event.rs
DevinR528: Yea that's actually what I had in the first commit :)

ruma-events-macros/src/event.rs
```diff
@@ -8,7 +8,7 @@ use syn::{Data, DataStruct, DeriveInput, Field, Fields, FieldsNamed, Ident};
 pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
     let ident = &input.ident;
     let (impl_gen, ty_gen, where_clause) = input.generics.split_for_impl();
-    let is_presence_or_redaction = ident == "PresenceEvent" || ident == "RedactionEvent";
+    let is_presence_or_redaction = input.generics.params.is_empty();
```
jplatte, line 11: There might be more event types in the future that are not generic. Could you either rename this to `is_generic` and "invert" all of the branches on it, or find another name that doesn't refer to presence and redaction events in particular?

ruma-events-macros/src/event.rs
```diff
@@ -8,7 +8,7 @@ use syn::{Data, DataStruct, DeriveInput, Field, Fields, FieldsNamed, Ident};
 pub fn expand_event(input: DeriveInput) -> syn::Result<TokenStream> {
     let ident = &input.ident;
     let (impl_gen, ty_gen, where_clause) = input.generics.split_for_impl();
-    let is_presence_or_redaction = input.generics.params.is_empty();
+    let is_generic = input.generics.params.is_empty();
```
jplatte, line 11: Now it does the opposite of what its name implies.

PR #120, opened 2020-06-09
Extend MembershipChange::ProfileChanged

jplatte:
Feature request by @dkasak.

@poljar, @iinuwa: Do you think this is sensible to have?


iinuwa:
Did a once-over, looks good to me

PR #122, opened 2020-06-10
Add trybuild tests for event_enum! macro

DevinR528:
Not sure if these are super helpful but might as well test what we can.


jplatte:
I think the first error message could be improved easily. The assert could have a message like 'well-known matrix events have to start with `m.`'.

